PR #5, opened 2020-01-10
Update dev dependency

ishche:


michalbali256:
Are we sure that the project can be compiled with these changes?

ishche:
@michalbali256  It worked for me. Do you experience any problems?

PR #14, opened 2020-03-17
fix: Fuzzer fixing

asmelko:
This PR fixes major bugs found by the parser library fuzzer.


parser_library/CMakeLists.txt
```diff
@@ -250,6 +250,11 @@ if(BUILD_FUZZER)
 
     set_target_properties(library_fuzzer PROPERTIES LINK_FLAGS "-fsanitize=fuzzer")
 	set_target_properties(library_fuzzer PROPERTIES COMPILE_FLAGS "-DANTLR4CPP_STATIC")
+
```
michalbali256: Is the file copied properly every time it is changed?

parser_library/fuzzer.dict
```diff
@@ -0,0 +1,41 @@
...
+"ORG"
+"L'"
+"I'"
+"S'"
+"T'"
+"D'"
+"O'"
+"N'"
+"K'"
+"C'ABC'"
```
michalbali256: Have you thought about adding some machine instructions as well?

parser_library/CMakeLists.txt
asmelko: If the input file is modified the build system will re-run CMake to re-configure the file and generate the build system again.

parser_library/fuzzer.dict
asmelko: Yes, I will add more extensive dictionary in future commits.

parser_library/src/context/ordinary_assembly/dependency_collector.cpp
```diff
@@ -121,14 +121,16 @@ dependency_collector& dependency_collector::div_mul(const dependency_collector&
...
+	auto unknown_space = std::find_if(addr.spaces.begin(), addr.spaces.end(), [](auto& entry) { return entry.first->kind == context::space_kind::LOCTR_UNKNOWN; });
 	if (unknown_space != addr.spaces.end())
 	{
-		for (auto it = addr.spaces.begin(); it != addr.spaces.end(); ++it)
-			if (it->first->kind != context::space_kind::LOCTR_UNKNOWN)
-				it->first->remove_listener(&addr);
-		auto entry = *unknown_space;
-		addr.spaces.clear();
-		addr.spaces.push_back(std::move(entry));
+		for (int i = addr.spaces.size() - 1; i > 0; --i)
```
slavek-kucera: is the zero-th element ignored intentionally here?

asmelko: no! I will fix it.

asmelko:
It is true that the above mentioned pairs would use static vector but for now I would try to retain the idea that each statement field is optional. I will address this problem in future PRs and will try to get rid of the pairs.

parser_library/src/context/ordinary_assembly/dependency_collector.cpp
```diff
@@ -121,14 +121,16 @@ dependency_collector& dependency_collector::div_mul(const dependency_collector&
 
 void dependency_collector::adjust_address(address& addr)
 {
-	auto unknown_space = std::find_if(addr.spaces.begin(), addr.spaces.end(), [] (auto& entry) { return entry.first->kind == context::space_kind::LOCTR_UNKNOWN; });
+	auto unknown_space = std::find_if(addr.spaces.begin(), addr.spaces.end(), [](auto& entry) { return entry.first->kind == context::space_kind::LOCTR_UNKNOWN; });
```
slavek-kucera, line 124: I suspect that the erase inside the loop causes O(N^2) behavior.
```suggestion
	auto known_spaces = std::partition(addr.spaces.begin(), addr.spaces.end(), [](auto& entry) { return entry.first->kind == context::space_kind::LOCTR_UNKNOWN; });
	std::for_each(known_spaces, addr.spaces.end(), [&addr](auto& entry) { return entry.first->remove_listener(&addr); });
	addr.spaces.erase(known_spaces, addr.spaces.end());
```

PR #16, opened 2020-03-25
Theia dap fix

michalbali256:


clients/vscode-hlasmplugin/src/extension.ts
```diff
@@ -29,6 +29,9 @@ const useTcp = false;
 var highlight: ASMSemanticHighlightingFeature;
```
ishche: This file is too big. Please consider refactoring.

language_server/src/dap/feature_launch.cpp
```diff
@@ -51,7 +51,7 @@ json feature_launch::register_capabilities()
 {
 	return json
 	{
-		{"supportsConfigurationDoneRequest", true}
+		{"supportsConfigurationDoneRequest", false}
```
ishche: are you sure that is the only capabilities that you support?
https://microsoft.github.io/debug-adapter-protocol/specification#Types_Capabilities

michalbali256:
this PR was to master, should go to development

PR #19, opened 2020-04-02
docs: readme update

michalbali256:


clients/vscode-hlasmplugin/README.md
```diff
@@ -123,17 +123,22 @@ The following example specifies that GROUP1 is used when working with `source_co
...
     },
-  ]
+  ],
+  "alwaysRecognize" : ["*.hlasm", "libs/*.asm"]
 }
 ```
 If you have the two configuration files configured as above and invoke the MAC1 macro from `source_code.hlasm`, the folder `ASMMAC/` in the current workspace is searched for a file with the exact name "MAC1". If that search is unsuccessful the folder `C:/SYS.ASMMAC` is searched. If that search is unsuccessful an error displays that the macro does not exist.
 
-The program field in `pgm_conf.json` supports regular expressions, for example:
+There is also option `alwaysRecognize` which takes array of wildcards. It allows you to configure two things:
```
asmelko, line 132:
```suggestion
There is also the option `alwaysRecognize` which takes an array of wildcards. It allows you to configure two things:
```

clients/vscode-hlasmplugin/README.md
```diff
@@ -123,17 +123,22 @@ The following example specifies that GROUP1 is used when working with `source_co
...
+- All files matching these wildcards will be always recognized as HLASM files. 
+- Macros and copy files will be recognized even if their file names does not match the identifier written into source code. For example, macro `IF` would be found even if it was defined in file `IF.hlasm`.
```
asmelko, line 134:
```suggestion
- Macro and copy files will be recognized even if their file names do not match the identifier used in a source code. For example, macro `IF` would be found even if it was defined in file `IF.hlasm`.
```

clients/vscode-hlasmplugin/README.md
```diff
@@ -123,17 +123,22 @@ The following example specifies that GROUP1 is used when working with `source_co
...
+
+The program field in `pgm_conf.json` supports wildcards, for example:
 ```
 {
   "pgms": [
     {
-      "program": ".*",
+      "program": "*",
```
asmelko, line 141: a bit vague example, don't you think?

clients/vscode-hlasmplugin/README.md
```diff
@@ -123,17 +123,22 @@ The following example specifies that GROUP1 is used when working with `source_co
...
-  ]
+  ],
+  "alwaysRecognize" : ["*.hlasm", "libs/*.asm"]
 }
 ```
 If you have the two configuration files configured as above and invoke the MAC1 macro from `source_code.hlasm`, the folder `ASMMAC/` in the current workspace is searched for a file with the exact name "MAC1". If that search is unsuccessful the folder `C:/SYS.ASMMAC` is searched. If that search is unsuccessful an error displays that the macro does not exist.
 
-The program field in `pgm_conf.json` supports regular expressions, for example:
+There is also option `alwaysRecognize` which takes array of wildcards. It allows you to configure two things:
+- All files matching these wildcards will be always recognized as HLASM files.
```
marcel1hruska, line 133:
```suggestion
- All files matching these wildcards will always be recognized as HLASM files. 
```

clients/vscode-hlasmplugin/README.md
michalbali256: that example should only demonstrate how to get the easiest setup. What do you suggest?

clients/vscode-hlasmplugin/README.md
marcel1hruska: This does not really explains the functionality. I would recommend to reformulate it, e.g. "If an extension wildcard is defined, all macro and copy files with such extension may be used in the source code. For example, with the extension wildcard '*.hlasm', a user may add macro __MAC__ to his source code even if it is in a file called __Mac.hlasm__."

clients/vscode-hlasmplugin/README.md
asmelko: maybe something that demonstrates wildcards more, `*.asm`. I agree with your response though, so it's OK for me.

marcel1hruska: common case would probably be a folder with open source files, e.g. pgms/*

michalbali256: what about "ASMPGM/*" ?

PR #20, opened 2020-04-02
Update issue templates

michalbali256:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
```
asmelko, line 18: Is the enumeration just a set of examples or fixed steps how to reproduce? If it is the set, it should not be numbered IMO.  Also I do not understand See error step.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
asmelko, line 11:
```suggestion
A clear and concise description of what the problem is. E.g., I'm always frustrated when [...]
```

.github/ISSUE_TEMPLATE/bug_report.md
michalbali256: It is just example. The last step means that once you make steps 1-3 then you can see the error.
BTW all of this is just template provided to me by GitHub.

PR #26, opened 2020-04-08
clang-format added and applied

michalbali256:


.clang-format
```diff
@@ -0,0 +1,49 @@
...
+  SplitEmptyRecord: false
+  SplitEmptyNamespace: false
+
+BreakBeforeBinaryOperators: NonAssignment
+FixNamespaceComments: true
+
+SortIncludes: true
+IncludeBlocks: Regroup
+IncludeCategories:
+  - Regex:           '<[[:alnum:]._]+>'
```
asmelko, line 30: I think that character `/` should be there as well.

.clang-format
```diff
@@ -0,0 +1,49 @@
...
+    Priority:        3
+
+IndentCaseLabels: true
+NamespaceIndentation: None
+IndentPPDirectives: AfterHash
+KeepEmptyLinesAtTheStartOfBlocks: false
+MaxEmptyLinesToKeep: 3
+
+SpaceAfterTemplateKeyword: false
+Standard: Cpp11
```
asmelko, line 44: what about Cpp17?

michalbali256: cpp17 is supported only since clang-format-10 and yields the same results as cpp11.

.clang-format
michalbali256: Is there any include from standard library that has '/' in it?

asmelko: seems that none

PR #32, opened 2020-04-19
Extension maintanance

marcel1hruska:
- added npm tests (also to pipeline)
- refactored extension


.github/workflows/alpine.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         run: mkdir build && cd build && cmake -G Ninja ../
       - name: Build
         run: cd build && cmake --build .
-      - name: Test
+      - name: Server Test
```
michalbali256, line 40: Is it possible on alpine to start at least those tests that do not require gui?
Or do all require gui?

clients/vscode-hlasmplugin/package.json
```diff
@@ -187,7 +188,7 @@
       "type": "object",
       "title": "hlasmplugin configuration",
       "properties": {
-        "semanticHighlightingColors": {
+        "hlasmplugin.semanticHighlightingColors": {
```
michalbali256, line 191: Are we still using name hlasmplugin? Maybe just hlasm or HlasmLanguageSupport

marcel1hruska: all other properties have that name. but simply 'hlasm' seems more appropriate

.github/workflows/alpine.yml
marcel1hruska: The whole testing suite requires gui. Thats the vscode way

michalbali256: OK. In the future, we should find a way to test it with theia & selenium

clients/vscode-hlasmplugin/.gitignore
```diff
@@ -1,4 +1,4 @@
 node_modules
 lib
 hlasmplugin.log
-bin
```
michalbali256, line 4: Why is bin removed? The contents of bin (language server executables) should never be checked into the git

marcel1hruska: This change for the build/bin folder, not bin/. Im not really sure why but it works, the build/bin files were updated and bin/ was not uploaded.

PR #34, opened 2020-04-21
chore: benchmark improvements

michalbali256:
benchmark now correctly continues when there is a parsing error with one file.
Refactoring of the benchmark.cpp file.

### DIFF
--- a/benchmark/benchmark.cpp
+++ b/benchmark/benchmark.cpp
@@ -98,7 +98,7 @@ struct all_file_stats
     size_t failed_file_opens = 0;
 };
 
-void one_file(const std::string & source_file, json& result, const std::string ws_folder, all_file_stats& s)
+void parse_one_file(const std::string& source_file, json& result, const std::string& ws_folder, all_file_stats& s)
 {
     auto source_path = ws_folder + "/" + source_file;
     std::ifstream in(source_path);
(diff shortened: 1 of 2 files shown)


benchmark/benchmark.cpp
```diff
@@ -87,6 +87,115 @@ class metrics_collector : public hlasm_plugin::parser_library::performance_metri
...
+    double average_line_ms = 0;
+    double average_stmt_ms = 0;
+    size_t all_files = 0;
+    long long whole_time = 0;
+    size_t program_count = 0;
+    size_t parsing_crashes = 0;
+    size_t failed_file_opens = 0;
+};
+
+void one_file(const std::string & source_file, json& result, const std::string ws_folder, all_file_stats& s)
```
asmelko, line 101: I suggest changing the method name to something that _says_ what is done with 'one file' eg. process/start_one_file

benchmark/benchmark.cpp
```diff
@@ -170,122 +290,44 @@ int main(int argc, char** argv)
...
-    {
-        std::clog << "Malformed json" << std::endl;
-        return 1;
-    }
+    std::clog << "Parse finished\n\n" << std::endl;
 
+    std::cout << result.dump(2) << std::endl;
 
     return 0;
 }
```
asmelko, line 333: newline

PR #51, opened 2020-05-20
test: Integration tests

marcel1hruska:
- Added a set of integration tests for VSCode and Theia


parser_library/src/context/ordinary_assembly/address.cpp
```diff
@@ -80,7 +80,7 @@ void space::resolve(space_ptr this_space, space_ptr value)
 
     assert(this_space->kind == space_kind::LOCTR_UNKNOWN);
 
-    for (int i = this_space->listeners_.size() - 1; i >= 0; --i)
+    for (int i = (int)(this_space->listeners_.size() - 1); i >= 0; --i)
```
michalbali256, line 83:
```suggestion
    for (int i = (int)this_space->listeners_.size() - 1; i >= 0; --i)
```

parser_library/src/context/ordinary_assembly/address.cpp
```diff
@@ -101,7 +101,7 @@ void space::resolve(space_ptr this_space, address value)
...
```
michalbali256, line 104:
```suggestion
    for (int i = (int)this_space->listeners_.size() - 1; i >= 0; --i)
```

PR #61, opened 2020-06-15
test: Improve test coverage of Language Server component

michalbali256:


language_server/src/request_manager.cpp
```diff
@@ -64,12 +70,8 @@ void request_manager::end_worker()
...
 bool hlasm_plugin::language_server::request_manager::is_running() 
 {
-    bool result = false;
-    {
-        std::unique_lock<std::mutex> lock(q_mtx_);
-        result = !requests_.empty();
-    }
-    return result;
+    std::lock_guard<std::mutex> lock(q_mtx_);
+    return !requests_.empty();
```
marcel1hruska, line 74: Just a note, I did exactly this in the past and for some reason it was breaking the server... As long as it works now.

michalbali256: Do you have any idea under what conditions it used to break? Because I am pretty sure these two implementations are equivalent.

CMakeSettings.json
```diff
@@ -93,6 +93,21 @@
       "addressSanitizerRuntimeFlags": "detect_leaks=0",
       "variables": []
     },
+    {
+      "name": "WSL-Clang-Release-ASAN",
+      "generator": "Ninja",
+      "configurationType": "Release",
+      "buildRoot": "${projectDir}\\out\\build\\${name}",
```
asmelko, line 100: can we unify build & install root dir for x64/86 and WSL?

language_server/src/request_manager.cpp
```diff
@@ -64,12 +70,8 @@ void request_manager::end_worker()
 
 bool hlasm_plugin::language_server::request_manager::is_running() 
 {
-    bool result = false;
-    {
-        std::unique_lock<std::mutex> lock(q_mtx_);
-        result = !requests_.empty();
-    }
-    return result;
+    std::lock_guard<std::mutex> lock(q_mtx_);
```
asmelko, line 73: i think that this lock is redundant. the whole method seems suspicious. ideally, you want to put lock in some if statement rather then in reading bool

```
{
std::lock_guard<std::mutex> lock(q_mtx_);
if (is_running())
   ...
}
```

language_server/src/request_manager.cpp
```diff
@@ -64,12 +70,8 @@ void request_manager::end_worker()
 
 bool hlasm_plugin::language_server::request_manager::is_running()
```
asmelko, line 71:
```suggestion
bool request_manager::is_running() 
```

CMakeSettings.json
michalbali256: where should be that unified location?

asmelko: the one for the x64/86 was OK IMO.

PR #63, opened 2020-07-09
fix: Absolute paths for processor groups

marcel1hruska:
Macros with extensions now work even for processor groups with absolute paths


michalbali256:
superseded by #117

PR #67, opened 2020-07-16
chore: Issue automation

asmelko:
Added new workflow so the new issues are automatically assigned to the Project Board.

### DIFF
--- a/.github/workflows/Issue.yml
+++ b/.github/workflows/Issue.yml
@@ -1,0 +1,29 @@
+#
+# Copyright (c) 2019 Broadcom.
+# The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+#
+# This program and the accompanying materials are made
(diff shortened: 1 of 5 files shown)


.github/workflows/Issue.yml
```diff
@@ -0,0 +1,29 @@
...
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Broadcom, Inc. - initial API and implementation
+#
+
+name: Move new issues into the project board
```
michalbali256, line 15: I suggest a shorter name for the action... It shows in the actions tab and I think the tab will be more organized with short names of the actions
```suggestion
name: Issue automation
```

Maybe rename the yml file to correspond to the workflow name as well.

PR #70, opened 2020-08-04
fix: EXRL format

marcel1hruska:


parser_library/src/context/instruction.cpp
```diff
@@ -571,7 +571,7 @@ hlasm_plugin::parser_library::context::instruction::get_machine_instructions()
     add_machine_instr(result, "EX", mach_format::RX_a, { reg_4_U, dxb_12_4x4_U }, 32, 740);
     add_machine_instr(result, "XIHF", mach_format::RIL_a, { reg_4_U, imm_32_S }, 48, 740);
     add_machine_instr(result, "XILF", mach_format::RIL_a, { reg_4_U, imm_32_S }, 48, 740);
-    add_machine_instr(result, "EXRL", mach_format::RIL_b, { reg_4_U, imm_32_S }, 48, 740);
+    add_machine_instr(result, "EXRL", mach_format::RIL_b, { reg_4_U, db_12_4_U }, 48, 740);
```
slavek-kucera, line 574: Doesn't `db_12_4_U` mean D12(B4) argument type?

Also do the `LARL`, `LGRL` and others suffer from the same issue?

michalbali256: Yes, it does.

The EXRL instruction (and probably others, if you say so) takes either relocatable symbol or an immediate value (with a warning). However in the machine code, the relocatable symbol is compiled into a relative address represented by a signed 32 bit immediate - that is where the mistake came from.

Unfortunately we do not support such format at this moment, so we want to make a hotfix now so when you use relocatable symbol with EXRL it does not report as false error. Then add an issue regarding the new instruction format.

Do you have more examples of such relative addressing instructions?

slavek-kucera: BPRP, BRAS, BRASL, BRC, BRCL, BRCT, BRCTG, BRCTH, BRXH, BRXHG, BRXLE, BRXLG, CGFRL, CGHRL, CGIJ, CGRJ, CGRL, CHRL, CIJ, CLGFRL, CLGHRL, CLGIJ, CLGRJ, CLGRL, CLHRL, CLIJ, CLRJ, CLRL, CRJ, CRL, EXRL, LARL, LGFRL, LGHRL, LGRL, LHRL, LLGFRL, LLGHRL, LLHRL, LRL, PFDRL, STGRL, STHRL, STRL

michalbali256:
We will solve the problem by introducing a new operand format in a new PR.

PR #86, opened 2020-09-15
fix: False positives

asmelko:
- Moved evaluation of ordinary symbols to the end of the assembly. Also, added default values to the undefined symbols to prevent unnecessary chained diagnostics.
- Changed resolution of `space` dependant (got rid of address listeners).
- Fixed another bunch of false positives.


parser_library/src/semantics/variable_symbol.cpp
```diff
@@ -27,10 +27,16 @@ basic_variable_symbol::basic_variable_symbol(
 
 context::id_index basic_variable_symbol::evaluate_name(const expressions::evaluation_context&) const { return name; }
 
+concat_chain&& clear_chain(concat_chain& chain)
```
slavek-kucera, line 30: This is a bit confusing... pick &, or &&.

slavek-kucera:
2d1ba17091cb2d58ec55535d1c739e5ae68d6313 reverted the JC alias.

asmelko:
Launched fuzzer several times. Added handling of empty smart pointers in grammar rules so now the handling code is not distributed over the whole project.

michalbali256 (approved):

PR #87, opened 2020-09-30
Semantic tokens

marcel1hruska:
Simulates upcoming LSP feature for semantic highlighting


clients/vscode-hlasmplugin/src/semanticTokens.ts
```diff
@@ -0,0 +1,151 @@
...
+
+	}
+
+	protected editorPredicate(uri: string): (editor: vscode.TextEditor) => boolean {
+		const predicateUri = vscode.Uri.parse(uri);
+		return (editor: vscode.TextEditor) => editor.document.uri.toString() === predicateUri.toString();
+	}
+
+	askForTokens(document: vscode.TextDocument) {
+		setTimeout(() => {
```
slavek-kucera, line 83: what does this do? Why the delay?

parser_library/src/parsing/parser_impl.cpp
```diff
@@ -375,6 +375,21 @@ void parser_impl::process_next(processing::statement_processor& proc)
 
 bool parser_impl::finished() const { return finished_flag; }
 
+void parser_impl::set_source_indices(const antlr4::Token* start, const antlr4::Token* stop)
+{
+    assert(stop);
+    size_t start_offset;
+
+    auto a1 = statement_start();
+    auto a2 = statement_end();
```
slavek-kucera, line 384: This was flagged by the sonarcloud and it really does look redundant.

PR #88, opened 2020-10-07
feat: Diagnostics suppression for files with no pgm_conf configuration

michalbali256:
Added the diagnosticsSuppressLimit option, that a number and limits the number of diagnostics that are shown when there is no configuration for currently opened file in the pgm_conf. The reason why to introduce such limit is that the majority of HLASM source files use macros and copy files. When there is no configuration, symbols defined in those external files are not accessible thus the plugin shows many diagnostics that are not useful. To increase usability of the plugin in such cases, we do not show diagnostics at all.

The limit allows to edit small files with small amount of diagnostics even without configuration. It can be configured either in VS Code settings for user-specific configuration or in pgm_conf for project-specific configuration. When both are configured, the pgm_conf is prioritized.

- Added handler for `workspace/didChangeConfiguration` notification
- Added implementation of LSP requests from server to client, needed for `workspace/configuration` request
- Added singleton class to parser_library that holds client configuration


michalbali256:
Closing as the I opened the PR to master  instaead of development by mistake

PR #89, opened 2020-10-07
feat: Diagnostics suppression for files with no pgm_conf configuration

michalbali256:
Added the diagnosticsSuppressLimit option, that a number and limits the number of diagnostics that are shown when there is no configuration for currently opened file in the pgm_conf. The reason why to introduce such limit is that the majority of HLASM source files use macros and copy files. When there is no configuration, symbols defined in those external files are not accessible thus the plugin shows many diagnostics that are not useful. To increase usability of the plugin in such cases, we do not show diagnostics at all.

The limit allows to edit small files with small amount of diagnostics even without configuration. It can be configured either in VS Code settings for user-specific configuration or in pgm_conf for project-specific configuration. When both are configured, the pgm_conf is prioritized.

- Added handler for `workspace/didChangeConfiguration` notification
- Added implementation of LSP requests from server to client, needed for `workspace/configuration` request
- Added a class to parser_library that holds client/workspace configuration


language_server/src/lsp/feature_workspace_folders.cpp
```diff
@@ -120,4 +132,28 @@ void feature_workspace_folders::did_change_watched_files(const json&, const json
...
+        std::bind(&feature_workspace_folders::configuration, this, std::placeholders::_1, std::placeholders::_2));
+    ++config_request_number;
+}
+
+void feature_workspace_folders::configuration(const json&, const json& params)
+{
+    if (params.size() == 0)
+        LOG_WARNING("Empty configuration response received.");
+
+    lib_config::load_from_json(params[0]);
```
slavek-kucera, line 154: does this make sence when `params.size() == 0`?

language_server/src/lsp/lsp_server.cpp
```diff
@@ -117,6 +167,17 @@ void server::on_initialize(json id, const json& param)
 
     respond(id, "", capabilities);
 
+    json register_configuration_changed_args {
+        { { "registrations", { { { "id", "configureRegister" }, { "method", "workspace/didChangeConfiguration" } } } } }
+    };
+    /*json register_message{ { "jsonrpc", "2.0" },
```
slavek-kucera, line 173: remove?

parser_library/src/workspaces/workspace.cpp
```diff
@@ -65,6 +67,36 @@ bool workspace::program_id_match(const std::string& filename, const program_id&
...
+{
+    if (message_consumer_)
+        message_consumer_->show_message(message, message_type::MT_INFO);
+}
+
+int64_t workspace::get_suppress_diags_limit()
+{
+    if (suppress_diags_limit_)
+        return suppress_diags_limit_.value();
+    return lib_config::get_instance()->diag_supress_limit;
```
slavek-kucera, line 97: shouldn't the lib_config instance be just another member of the workspace?

language_server/src/lsp/feature_workspace_folders.cpp
```diff
@@ -120,4 +132,28 @@ void feature_workspace_folders::did_change_watched_files(const json&, const json
         paths.begin(), paths.end(), std::back_inserter(c_uris), [](const std::string& s) { return s.c_str(); });
     ws_mngr_.did_change_watched_files(c_uris.data(), c_uris.size());
 }
+
+void feature_workspace_folders::send_configuration_request()
+{
+    if (!response_)
+        return;
+    static const json config_request_args { { "items", { { { "section", "hlasm" } } } } };
+    static uint64_t config_request_number = 0;
```
slavek-kucera, line 141: Any concerns regarding access from multiple threads?

Do the numbers need to be only unique or do they actually have to form an uninterrupted sequence?

michalbali256: I made the variable a member variable, there was really no reason to make it static...

The numbers just need to be unique.

language_server/src/lsp/feature_workspace_folders.cpp
michalbali256: No, it does not, thanks.

parser_library/src/workspaces/workspace.cpp
michalbali256: The configuration can be changed by the user during lifetime of workspace, so we need to get the instance every time.

parser_library/src/lib_config.cpp
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Broadcom, Inc. - initial API and implementation
+ */
+
+#include "lib_config.h"
+
+static std::shared_ptr<lib_config> config_instance = std::make_shared<lib_config>();
```
asmelko, line 17: is static needed here?

parser_library/src/workspaces/workspace.cpp
```diff
@@ -142,6 +173,18 @@ void workspace::parse_file(const std::string& file_uri)
         f->parse(*this);
         if (!f->dependencies().empty())
             dependants_.insert(f->get_file_name());
+
+
+        // if there is no processor group assigned to the program, delete diagnostics that may have been created
+        if (cancel_ && cancel_->load()) // skip, if parsing was cancelled using the cancellation token
```
asmelko, line 179: can you simplify?
```suggestion
        if (cancel_ && *cancel_) // skip, if parsing was cancelled using the cancellation token
```

parser_library/include/workspace_manager.h
```diff
@@ -114,7 +115,7 @@ class PARSER_LIBRARY_EXPORT workspace_manager
     virtual void register_highlighting_consumer(highlighting_consumer* consumer);
     virtual void register_diagnostics_consumer(diagnostics_consumer* consumer);
     virtual void register_performance_metrics_consumer(performance_metrics_consumer* consumer);
-
+    virtual void set_message_consumer(message_consumer* consumer);
```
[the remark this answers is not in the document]
asmelko: or, does it make sense to register more message consumers?

parser_library/src/workspaces/workspace.cpp
slavek-kucera: I was thinking that perhaps each workspace might have different limits?

parser_library/src/lib_config.cpp
slavek-kucera: could be anonymous namespace instead...

But I am still wondering if this should be part of the server's state.

parser_library/src/workspaces/workspace.cpp
michalbali256: Yes, right now the static lib_config instance represents the configuration of VS Code client. At the same time, you can override the limit of each workspace by writing the configuration into pgm_conf.json.

michalbali256: I can move the static instance to workspace_manager, would that be preferable?
But we still need some code to merge the global and workspace configurations.

slavek-kucera: I think that you should not have a static instance at all.

michalbali256: Sorry for late editing of the previous comment.

I meant adding the instance as a non-static member of workspace manager.

asmelko (approved):

language_server/src/server.h
```diff
@@ -58,6 +58,7 @@ class server : public response_provider
     std::vector<std::unique_ptr<feature>> features_;
 
     std::map<std::string, method> methods_;
+    std::unordered_map<json, method> request_handlers_;
```
slavek-kucera, line 61: could you elaborate on why use `json` as the key?

PR #106, opened 2021-02-15
US704269 | Provide user ability to specify compiler options

SWETAS04:
Implementation of assembler options "SYSPARM"  from proc_grps.json


clients/vscode-hlasmplugin/proc_grps_schema
```diff
@@ -7,17 +7,32 @@
...
-                            "type" : "string"
+                        "type": "array",
+                        "description": "List of folders containing external files such as macros and copy.\nAll files inside these folders are being automatically recognized as HLASM.\nIf you would like to use dependency files with a specific extension, create an extension wildcard in pgm_conf.json for it.\nExtension wildcards must follow the format: [anything]*.[1+ non-whitespace characters], e.g. libs/*.hlasm or *.asm",
+                        "items": {
+                            "type": "string"
                         }
+                    },
+                    "asm_options": {
+                        "type": "object",
+                        "description": "List of Assembler Compile",
```
slavek-kucera, line 24:
```suggestion
                        "description": "List of assembler options",
```

clients/vscode-hlasmplugin/proc_grps_schema
```diff
@@ -7,17 +7,32 @@
...
+                            "SYSPARM": {
+                                "type": "string",
+                                "description": "It specifies the character string the assembler assigns to the &SYSPARM system variable symbol. "
+                            },
+                            "PROFILE": {
+                                "type": "string",
+                                "description": "Profile Member to be copied int the source program."
+                            }
+                        },
+                        "required": [ "SYSPARM", "PROFILE" ]
```
slavek-kucera, line 35: I don't think these should be required.

example_workspace/.hlasmplugin/proc_grps.json
```diff
@@ -1,10 +1,15 @@
...
-        "libs","libs2"
-      ]
+        "name": "P1",
+        "libs": [
+            "libs",
+            "libs2"
+        ],
+        "asm_options": {
+            "SYSPARM": "STH1",
+            "PROFILE": "STH2"
```
slavek-kucera, line 11: I'd prefer if the example would actually made sense in the context of the sample workspace - so ideally a real profile macro that would utilize the `SYSPARM` value?

parser_library/src/context/hlasm_context.cpp
```diff
@@ -160,7 +160,8 @@ void hlasm_context::add_global_system_vars()
     auto SYSDATC = ids().add("SYSDATC");
     auto SYSDATE = ids().add("SYSDATE");
     auto SYSTIME = ids().add("SYSTIME");
-    auto SYSPARM = ids().add("SYSPARM");
+    auto SYSPARM = asm_options_.find("SYSPARM") == asm_options_.end() ? ids().add("SYSPARM")
```
slavek-kucera, line 163: This seems wrong. I suspect that what this code does is just getting the name "SYSPARM" for the variable and not the value yet. I think that happens at `auto val = std::make_shared<set_symbol<C_t>>(SYSPARM, true, true);`

parser_library/src/context/hlasm_context.cpp
```diff
@@ -248,12 +249,13 @@ bool hlasm_context::is_opcode(id_index symbol) const
...
 }
 
-hlasm_context::hlasm_context(std::string file_name)
+hlasm_context::hlasm_context(std::string file_name, std::map<std::string, std::string> asm_options)
     : instruction_map_(init_instruction_map())
     , SYSNDX_(0)
     , ord_ctx(ids_)
     , lsp_ctx(std::make_shared<lsp_context>())
 {
+    asm_options_ = asm_options;
```
slavek-kucera, line 258:
```suggestion
    asm_options_ = std::move(asm_options);
```
or better yet, move it to the initialization part.

parser_library/src/debugging/debug_lib_provider.h
```diff
@@ -57,6 +57,12 @@ class debug_lib_provider : public workspaces::parse_lib_provider
 
         return false;
     }
+    std::map<std::string, std::string> get_asmOptions(const std::string& file_name)
```
slavek-kucera, line 60: Probably keep the lower case style - `get_asm_options`

parser_library/src/workspaces/processor_group.h
```diff
@@ -38,13 +38,22 @@ class processor_group : public diagnosable_impl
 
     void add_library(std::unique_ptr<library> library) { libs_.push_back(std::move(library)); }
 
+    void add_asm_options(std::vector<std::string> asm_options)
+    {
+        asm_options_.insert(make_pair("SYSPARM", asm_options.at(0)));
+        asm_options_.insert(make_pair("PROFILE", asm_options.at(1)));
+    }
+
```
slavek-kucera, line 46: I really don't like this interface... It is not obvious that the only correct vector is the one with two elements, no way to tell easily at call site that the first one is sysparm, the other profile...

slavek-kucera (changes requested):

clients/vscode-hlasmplugin/proc_grps_schema
SWETAS04: I added it thinking if some one adds asm_options(which is optional ) , the user wants to use them.

example_workspace/.hlasmplugin/proc_grps.json
SWETAS04: Yes, I agree. I copied it simply from the story created in rally.

parser_library/src/context/hlasm_context.cpp
SWETAS04: It will try to find the value for the key "SYSPARM" and if not found will simply add "SYSPARM" else the value found in asm_options

parser_library/src/workspaces/processor_group.h
SWETAS04: Yes , you are right. Will change it to map to with key values "SYSPARM" and "Profile"

clients/vscode-hlasmplugin/proc_grps_schema
slavek-kucera: Yes, but for example I've never used the `PROFILE` parameter. So it makes sense to allow not specifying all of them.

parser_library/src/context/hlasm_context.cpp
slavek-kucera: But the `SYSPARM` option is always called `SYSPARM` and as far as I can tell that's what this code is supposed to do.

PR #108, opened 2021-02-18
feat: Provide users ability to use compiler option "SYSPARM"

SWETAS04:
Includes processing of SYSPARM option for the source code being compiled.


parser_library/src/workspaces/workspace.cpp
```diff
@@ -46,7 +46,9 @@ workspace::workspace(const ws_uri& uri,
 workspace::workspace(
     const ws_uri& uri, file_manager& file_manager, const lib_config& global_config, std::atomic<bool>* cancel)
     : workspace(uri, uri, file_manager, global_config, cancel)
-{}
+{
+    opened_ = true;
```
slavek-kucera, line 50: This does not seem right, a lot of thigs happen when `open()` is called.

parser_library/src/workspaces/parse_lib_provider.h
```diff
@@ -37,6 +37,8 @@ class parse_lib_provider
 
     virtual bool has_library(const std::string& library, context::hlasm_context& hlasm_ctx) const = 0;
 
+    virtual std::map<std::string, std::string> get_asm_options(const std::string&) = 0;
```
slavek-kucera, line 40: The more I think about this, the more unsure I am as to why to add this new member function into this particular interface.

slavek-kucera (changes requested):
I don't see any new unit tests.

parser_library/src/workspaces/parse_lib_provider.h
SWETAS04: It was added, to pass the SYSPARM value in the hlasm_context() constructor, dint get a better way to directly access workpace .cpp

slavek-kucera: There is quite a few parameters on the analyzer already. I am wondering whether a new struct `analyzer_options` should be added. And that would contain collect_hl_info, tracer, library_data and a new members related to sysparm and profile parameters.
What do you think @michalbali256 ?

michalbali256: I thought about this too and I think it is a good place to add the function. Moving forward, the interface should become more like "processor group provider" instead of just "library provider".

Currently, I do not even see a reason why should the `workspace` implement the interface instead of `workspaces::processor_group`, since every method first finds the correct processor group (that could be found just once when the parsing starts) and then uses members of the processor group to provide the library. But I see that as a refactor for another time.

example_workspace/.hlasmplugin/proc_grps.json
```diff
@@ -1,10 +1,14 @@
...
-      "libs": [
-        "libs","libs2"
-      ]
+       "name": "P1",
+       "libs": [
+          "libs","libs2"
+        ],
+        "asm_options": {
+            "SYSPARM": "SEVEN",
+            "PROFILE": "MAC1"
```
michalbali256, line 10:
```suggestion
```
I would remove the PROFILE option, since it is not implemented yet. And even when it is implemented, we do not need to showcase every single feature of the plugin in `example_workspace`... A PROFILE option set for all the simple example sources could be very confusing for newcomers.

parser_library/src/workspaces/parse_lib_provider.h
michalbali256: I couldn't see your latest answer as I was typing mine. Yes, it is a good idea for the future, since it is easier to manage a struct than several constructors with every change. However I still think adding the new virtual function into the `parse_lib_provider` is a good change

parser_library/src/workspaces/workspace.cpp
```diff
@@ -319,6 +322,20 @@ bool workspace::load_and_process_config()
                 // else ignore, publish warning
             }
         }
+        if (asm_options_json.size() != 0)
+        {
+            std::map<std::string, std::string> asm_options;
+
+            if (asm_options_json.count("SYSPARM"))
```
michalbali256, line 329: I am thinking that maybe we should just copy all key-string pairs from the json dictionary to the map? That way, adding new options in the future will not require change in this code.

parser_library/src/workspaces/workspace.cpp
```diff
@@ -319,6 +322,20 @@ bool workspace::load_and_process_config()
...
+            {
+                asm_options.insert({ "SYSPARM", asm_options_json["SYSPARM"].get<std::string>() });
```
michalbali256, line 331: will this code work when the user enters something different as string to the SYSPARM value?
e.g.
```
"asm_options" : {
  "SYSPARM" : 42
}
```

slavek-kucera: I think this violates the schema, so I wouldn't see it as an issue.
On the other hand, we may want to impose the 255 characters limit imposed in the doc.

parser_library/src/workspaces/workspace.cpp
slavek-kucera: Additional parameters may need validation etc.

parser_library/src/workspaces/parse_lib_provider.h
slavek-kucera: Will the function still make sense if we are able to pass the options directly to the analyzer?

michalbali256: I think so. A new parameter would require a change in processor_file_impl::parse signature to take one more parameter excapt `parse_lib_provider`. Also when we will implement additional options like profile, it will be possible to get them from deeper layers of analyzer which already have reference to `parse_lib_provider`

If there will be additional options not tied with processor groups that we need to pass into analyzer, we should revisit this conversation in the future.

parser_library/src/workspaces/workspace.cpp
michalbali256: It throws and then the exception is caught in `lsp::server`. It would be cleaner solution if we handled it directly in the configuration loading function

parser_library/src/workspaces/parse_lib_provider.h
slavek-kucera: So far I only see two callers of `parse`, both in workspace...

parser_library/include/workspace_manager.h
```diff
@@ -130,7 +130,11 @@ class PARSER_LIBRARY_EXPORT workspace_manager
 private:
     impl* impl_;
 };
-
+struct asm_option
```
michalbali256, line 133: this struct should be defined somewhere else... Workspace manager contains the interface between `parser_library` and `language_server`.
It should probably be defined in its separate file inside `hlasm_plugin::parser_library` namespace (so in `parser_library/src/` folder)

parser_library/src/context/hlasm_context.h
```diff
@@ -19,13 +19,15 @@
 #include <memory>
 #include <set>
 #include <vector>
+#include <workspace_manager.h>
```
michalbali256, line 22: context should definitely not depend on workspace manager

clients/vscode-hlasmplugin/proc_grps_schema
```diff
@@ -15,7 +15,7 @@
                     "libs": {
                         "type" : "array",
                         "description" : "List of folders containing external files such as macros and copy.\nAll files inside these folders are being automatically recognized as HLASM.\nIf you would like to use dependency files with a specific extension, create an extension wildcard in pgm_conf.json for it.\nExtension wildcards must follow the format: [anything]*.[1+ non-whitespace characters], e.g. libs/*.hlasm or *.asm",
-                        "items": {
+                         "items": {
```
michalbali256, line 18:
```suggestion
                        "items": {
```

michalbali256 (changes requested):

parser_library/src/workspaces/processor_group.h
```diff
@@ -38,14 +37,23 @@ class processor_group : public diagnosable_impl
 
     void add_library(std::unique_ptr<library> library) { libs_.push_back(std::move(library)); }
 
+    void add_asm_options(std::map<std::string, std::string> asm_options)
```
michalbali256, line 40:
```suggestion
    void add_asm_options(const std::map<std::string, std::string>& asm_options)
```

michalbali256 (approved):

PR #112, opened 2021-03-30
Feature : Support missing instructions 

SWETAS04:
This PR includes new instruction mnemonics and the modified ones that have been added to the new Z15 and support for missing ones in the previous version.


SWETAS04:
Wrong base

PR #113, opened 2021-03-30
feat: Support missing instruction

SWETAS04:
This PR includes new instruction mnemonics and the modified ones that have been added to the new Z15 and support for missing ones in the previous version.


parser_library/src/context/instruction.cpp
```diff
@@ -445,1252 +443,1396 @@ hlasm_plugin::parser_library::context::instruction::get_machine_instructions()
...
         mach_format::VRR_d,
         { vec_reg_4_U, vec_reg_4_U, vec_reg_4_U, vec_reg_4_U, mask_4_U, mask_4_U },
-        48,
+
         1573);
-    add_machine_instr(result, "VNN", mach_format::VRR_c, { vec_reg_4_U, vec_reg_4_U, vec_reg_4_U }, 48, 1574);
-    add_machine_instr(result, "VNO", mach_format::VRR_c, { vec_reg_4_U, vec_reg_4_U, vec_reg_4_U }, 48, 1574);
+    add_machine_instr(result, "VNN", mach_format::VRR_c, { vec_reg_4_U, vec_reg_4_U, vec_reg_4_U }, 1574);
+    add_machine_instr(result, "VNO", mach_format::VRR_c, { vec_reg_4_U, vec_reg_4_U, vec_reg_4_U }, 1574);
     result.insert(std::pair<const std::string, machine_instruction_ptr>("VNOT",
```
slavek-kucera, line 1559: I am wondering about this mnemonic. As far as I can tell VNOT has two operands VNOT A,B which is expanded into VNO A,B,B instruction. Yet the plugin requires three arguments at the moment.

SWETAS04: Yes, you are correct. But , this was something pre-existing in the plugin where it takes 3 arguments and wants 2nd and 3rd operands to be equal. I am not sure why it was done this way ?
Maybe @michalbali256 remembers something about it

michalbali256: So VNOT is actually a mnemonic? It must have been originally misunderstood... We can either keep VNOT as an instruction and just edit its format, or create a mnemonic - slight problem being, that we do not have a way to copy second operand and use it as third. However, we can just substitute third operand with zero, which will always be valid. At the same time, the second operand will still be checked so the user will not know.

Either way, class `vnot_instruction` can be removed.

Issue #4, opened 2020-01-09
Add possibility to search macro names in lowercase and with extension


JeffyOLOLO:
Hello!

In our project we store macros in lowercase and with extension _.sm_.
E.g. strucs/if.sm; maclib/@wto.sm

Is it possible to search for macros in this case? If not, will it be difficult to add?

Thank you!

michalbali256:
Hi!

Thank you for your feedback!

At the moment, it is only possible to use macros that are in a file that is named exactly same as the macro.

It should not be too hard to add to a future release.

marcel1hruska:
Hi!

Today we released version 0.10.0, which supports your use case.

We added an `alwaysRecognize` array of wildcards to the `pgm_conf.json`. If you define a wildcard that represents a custom extension (e.g. `"*.sm"`), you may use macro and copy files with this extension. Furthermore, we do not require these files to be upper-case.

JeffyOLOLO:
Hello, I updated the plugin and configured as you adviced but it doesn't work :(
For example, we use maclib/if.sm, maclib/endif.sm, maclib/@wto.sm, in source code it looks like:
```
       IF (FLAG1)
            @WTO   'SOMETHING'
       ENDIF
```
My proc_grps.json:
```
{
  "pgroups": [
    {
      "name": "macros",
      "libs": [
        "C:/Users/ ... /macapi",
        "C:/Users/ ... /maclib"
      ]
    }
  ]
}
```
pgm_conf.json:
```
{
  "pgms": [
    {
      "program": "C:/Users/ ... /asm/*",
      "pgroup": "macros"
    }
  ],
  "alwaysRecognize" : [
    "C:/Users/ ... /macapi/*.sm",
    "C:/Users/ ... /maclib/*.sm"
  ]
}
```
There are absolute path but it doesn't work with relative path too.

michalbali256:
I am sorry but I cannot recreate your problem.

The paths should be relative to your workspace.
Are you sure that you are using the 0.10.0 release? Also, what operating system do you use?

[example_ws.zip](https://github.com/eclipse/che-che4z-lsp-for-hlasm/files/4450371/example_ws.zip): here is an example workspace that works on my machine.

If you will still experience the problem can you wrap your workspace in a zip and attach it?

JeffyOLOLO:
I have Windows 10.0.17134
Unfortunatly it is working now on workspace level...
[...]

PR #117, opened 2021-04-08
feat: Support for macro file extensions

slavek-kucera:
The PR expands the `proc_conf.json` configuration file by adding a new `macro_extensions` section that could be used to provide a list of extensions to be used when search the macro libraries.
The `alwaysRecognize` in `pgm_conf.json` is deprecated, but a compatibility mechanism is implemented that extracts the extensions from the list and uses them as the default for `macro_extensions` in the `proc_conf.json`.

Fixes: #4


parser_library/src/workspaces/local_library.cpp
```diff
@@ -0,0 +1,159 @@
...
+            break;
+        case hlasm_plugin::utils::path::list_directory_rc::not_a_directory:
+            add_diagnostic(diagnostic_s::error_L0002(lib_path_));
+            break;
+        case hlasm_plugin::utils::path::list_directory_rc::other_failure:
+            add_diagnostic(diagnostic_s::error_L0001(lib_path_));
+            break;
+    }
+
+    bool extension_removed = true;
```
slavek-kucera, line 123: false!

parser_library/src/workspaces/workspace.cpp
```diff
@@ -327,134 +331,77 @@ bool workspace::load_and_process_config()
...
+            lib_path = utils::path::lexically_normal(lib_path);
+
+            library_local_options opts;
+            opts.optional_library = lib.optional;
+            if (!lib.macro_extensions.empty())
+                opts.extensions = lib.macro_extensions;
+            else if (!proc_groups.macro_extensions.empty())
+                opts.extensions = proc_groups.macro_extensions;
+            else if (!macro_extensions_compatibility_list.empty())
+                opts.extensions = macro_extensions_compatibility_list, opts.extensions_from_deprecated_source = true;
```
michalbali256, line 383:
```suggestion
{
                opts.extensions = macro_extensions_compatibility_list;
                opts.extensions_from_deprecated_source = true;
}
```

michalbali256 (commented):
Should there be any changes in clients/vscode-hlasmplugin/src/configurationsHandler.ts:126-137 now, that there can be a json object instead of string only? I cannot think of any remaining places where the configs are parsed right now.

Regarding code coverage - it is OK, but I suggest at least covering code paths that emit new diagnostics.

PR #121, opened 2021-04-19
fix: EXLR flagged as error by plugin

SWETAS04:
This PR incudes a new operand type which accepts either a relocatable symbol or absolute immediate value( with a warning). 
 Other Instructions which suffer from the same problem are also changed.


parser_library/src/diagnostic.cpp
```diff
@@ -1477,6 +1474,14 @@ diagnostic_op diagnostic_op::warn_D025(const range& range, const std::string& ty
     return diagnostic_op(
         diagnostic_severity::warning, "D025", "The " + modifier + " modifier is ignored with type " + type, range);
 }
+diagnostic_op diagnostic_op::warn_D031(const range& range, const std::string& instr)
+{
+    return diagnostic_op(diagnostic_severity::warning,
+        "D031",
+        " Operand value must be relocatable symbol with instruction " + instr,
```
slavek-kucera, line 1481: I think it would be more useful to report the operand we're generating the warning for instead of the instruction name + replace must with should as this is a warning.

parser_library/src/diagnostic.h
```diff
@@ -439,6 +439,7 @@ struct diagnostic_op
     static diagnostic_op error_D023(const range& range);
     static diagnostic_op error_D024(const range& range, const std::string& type);
     static diagnostic_op warn_D025(const range& range, const std::string& type, const std::string& modifier);
+    static diagnostic_op warn_D031(const range& range, const std::string& modifier);
```
slavek-kucera, line 442: reorder, please.

parser_library/src/diagnostic.h
```diff
@@ -457,20 +458,20 @@ struct diagnostic_op
 
     static diagnostic_op error_M112(const std::string& instr_name, const range& range);
 
-    static diagnostic_op error_M113(const std::string& instr_name, const range& range);
```
slavek-kucera, line 460: I think we should keep the original message ids, the messages relate to basically the same issues as before.

parser_library/src/checking/instr_operand.cpp
```diff
@@ -287,6 +287,13 @@ bool one_operand::check(
         }
         return true;
     }
+    // simple operand with relocatable symbol or immediate value
+
+    if (to_check.identifier.type == machine_operand_type::RELOC_IMM && operand_identifier != "RELOC")
```
slavek-kucera, line 292: This is basically reusing the `operand_identifier` as an flag, right? I suggest introducing that flag instead.

parser_library/test/diagnostics_check_test.cpp
```diff
@@ -36,7 +36,7 @@ TEST(diagnostics, overall_correctness)
 
     ASSERT_EQ(a.parser().getNumberOfSyntaxErrors(), (size_t)0);
 
-    ASSERT_EQ(dynamic_cast<hlasm_plugin::parser_library::diagnosable*>(&a)->diags().size(), (size_t)0);
+    ASSERT_EQ(dynamic_cast<hlasm_plugin::parser_library::diagnosable*>(&a)->diags().size(), (size_t)1);
```
slavek-kucera, line 39:
```suggestion
    ASSERT_EQ(dynamic_cast<hlasm_plugin::parser_library::diagnosable*>(&a)->diags().size(), (size_t)0);
```

I suggest fixing the source instead.

parser_library/test/diagnostics_check_test.cpp
```diff
@@ -236,7 +236,7 @@ TEST(diagnostics, mnemonics)
 
     ASSERT_EQ(a.parser().getNumberOfSyntaxErrors(), (size_t)0);
 
-    ASSERT_EQ(dynamic_cast<hlasm_plugin::parser_library::diagnosable*>(&a)->diags().size(), (size_t)0);
+    ASSERT_EQ(dynamic_cast<hlasm_plugin::parser_library::diagnosable*>(&a)->diags().size(), (size_t)20);
```
slavek-kucera, line 239:
```suggestion
    ASSERT_EQ(dynamic_cast<hlasm_plugin::parser_library::diagnosable*>(&a)->diags().size(), (size_t)0);
```

The same here, we should be able to create jump targets that are far enough.

parser_library/src/diagnostic.h
slavek-kucera: Please add a test generating this message.

slavek-kucera (changes requested):

parser_library/src/checking/instr_operand.cpp
```diff
@@ -297,8 +304,8 @@ bool one_operand::check(
             case machine_operand_type::IMM:
                 diag = diagnostic_op::error_M122(instr_name, -boundary, boundary - 1, operand_range);
                 break;
-            case machine_operand_type::REG_IMM:
-                diag = diagnostic_op::error_M123(instr_name, -boundary, boundary - 1, operand_range);
+            case machine_operand_type::RELOC_IMM:
```
michalbali256, line 307: Coupled with your other change in `operand_impls.cpp`, this code checks, whether the actual offset of relocatable symbol from the beginning of current section is within boundaries (function `is_size_corresponding_signed`). However that is wrong, the instructions we are fixing here take the difference between the relocatable symbol and the instruction adress and that should be checked for the right size.
Example: 
```
SECT CSECT
...   CODE1, which is very long
RELOC ...
... CODE2, shorter
  EXRL 1,RELOC
```
Currently, you are checking the length of CODE1 - the absolute offset from the beginning of section. However, the length of CODE2 should be checked - relative length between address of EXRL and RELOC.

To do this, we could probably introduce a special operand type whose value would be `*-<value>`. We would have to identify that we are generating specific operand for specific machine instructions...

Or extend checker API with the address of current checked instruction (however here I am not sure whether we know addresses of all instructions that are checked).

Or just remove the check and leave it for the future.

parser_library/src/semantics/operand_impls.cpp
```diff
@@ -90,17 +90,15 @@ std::unique_ptr<checking::operand> make_check_operand(expressions::mach_evaluate
     {
         return std::make_unique<checking::one_operand>(res.get_abs());
     }
+    else if (res.value_kind() == context::symbol_value_kind::RELOC && type_hint
+        && *type_hint == checking::machine_operand_type::RELOC_IMM)
+    {
+        return std::make_unique<checking::one_operand>("RELOC", res.get_reloc().offset());
```
michalbali256, line 96: This code would accept instructions like:
```
RELOC LR 1,1
 EXRL 1,RELOC+5
 EXRL 1,RELOC+RELOC
```
The second one is definitely invalid, I am not sure about the first one.
We should generate this special operand only when the relocatable symbol is not complex (`res.get_reloc().is_simple()`) and maybe even just when it is just a simple symbol (not sure about the latter).

slavek-kucera: the `exrl` takes the difference in halfwords (multiples of 2 bytes), so `reloc+5` cannot be represented at all.

michalbali256: OK, but `RELOC+4` would be fine?

slavek-kucera: yes, `reloc+4` is ok.

SWETAS04: Length being multiple of 2 bytes holds true only for exrl instruction or for all instructions with relative addressing?

slavek-kucera: as far as I know, it's for all reladdr operands.

parser_library/src/checking/instr_operand.cpp
SWETAS04: Yes, agreed. Even I was not sure if it would give the correct difference. I think (*- <value>) for specific machine instructions would be okay for now.

parser_library/src/diagnostic.cpp
```diff
@@ -1477,6 +1477,14 @@ diagnostic_op diagnostic_op::warn_D025(const range& range, const std::string& ty
     return diagnostic_op(
         diagnostic_severity::warning, "D025", "The " + modifier + " modifier is ignored with type " + type, range);
 }
+diagnostic_op diagnostic_op::warn_D031(const range& range, const std::string& operand_value)
+{
+    return diagnostic_op(diagnostic_severity::warning,
+        "D031",
+        " Operand value " + operand_value + " should be relocatable symbol. ",
+
+        range);
```
slavek-kucera, line 1486:
```suggestion
        "Operand value " + operand_value + " should be a relocatable symbol",
        range);
```

parser_library/src/expressions/mach_operator.h
```diff
@@ -137,7 +142,24 @@ inline mach_expression::value_t mach_expr_binary<sub>::evaluate(mach_evaluate_in
...
+    {
+        add_diagnostic(diagnostic_op::warn_D031(get_range(), std::to_string(right_->evaluate(info).get_abs())));
+        return right_->evaluate(info);
+    }
+    else
+    {
+        if (lc.get_reloc().offset() > symb.get_reloc().offset())
+            return (left_->evaluate(info) - right_->evaluate(info));
+        else
+            return (right_->evaluate(info) - left_->evaluate(info));
```
slavek-kucera, line 160: I don't understand this part... relative address may very well be positive or negative.
I'd expect simply `target-location`.

SWETAS04: Okay, thanks for pointing out. I misunderstood it as difference should be always positive value, will change it to target - location.

parser_library/src/parsing/parser_impl.cpp
```diff
@@ -543,24 +543,24 @@ void parser_impl::transform_imm_reg_operands(semantics::collector& col, id_index
...
         instruction_name = context::instruction::mnemonic_codes.at(*instruction).instruction->instr_name;
         replaced = context::instruction::mnemonic_codes.at(*instruction).replaced;
     }
     int position = 0;
-    for (auto& operand : *opernds)
+    for (const auto& operand : *opernds)
     {
         // replaced operands are skipped being immediate values
-        if (replaced.size() != 0 && replaced.at(position).first == position)
+        if (!replaced.empty() && replaced.at(position).first == position)
```
slavek-kucera, line 556: I think this is wrong... The `replaced` vector contains `index, value` pairs, so if the replacement occurs at position 3, it contains a single pair `3,value`, so `.at(3)` will throw.
Also there may be multiple replacements in a row, so this should probably be some kind of loop.

@michalbali256 - do you concur?

parser_library/src/parsing/parser_impl.cpp
```diff
@@ -543,24 +543,24 @@ void parser_impl::transform_imm_reg_operands(semantics::collector& col, id_index
...
     std::vector<std::pair<size_t, size_t>> replaced;
     auto opernds = &col.current_operands().value;
-    auto mnem_tmp = context::instruction::mnemonic_codes.find(*instruction);
-    if (mnem_tmp != context::instruction::mnemonic_codes.end())
+    if (auto mnem_tmp = context::instruction::mnemonic_codes.find(*instruction);
+        mnem_tmp != context::instruction::mnemonic_codes.end())
     {
         instruction_name = context::instruction::mnemonic_codes.at(*instruction).instruction->instr_name;
         replaced = context::instruction::mnemonic_codes.at(*instruction).replaced;
     }
```
slavek-kucera: This could be simplified a bit to something like
```suggestion
    const machine_instruction* instr;
    const std::vector<std::pair<size_t, size_t>>* replaced = nullptr;
    if (auto mnem_tmp = context::instruction::mnemonic_codes.find(*instruction);
        mnem_tmp != context::instruction::mnemonic_codes.end())
    {
        instr = mnem_tmp->instruction;
        replaced = &mnem_tmp->replaced;
    }
    else
    {
        instr = &context::instruction::machine_instructions.at(*instruction);
    }
```
but it needs to be properly incorporated into the rest of the code

parser_library/src/parsing/parser_impl.cpp
```diff
@@ -537,11 +541,12 @@ void parser_impl::process_lookahead()
             parse_lookahead_operands(std::move(*look_lab_instr->op_text), look_lab_instr->op_range);
     }
 }
-void parser_impl::transform_imm_reg_operands(semantics::collector& col, id_index instruction)
+void parser_impl::transform_imm_reg_operands(semantics::operand_list& op_list, id_index instruction)
 {
+    if (instruction->empty())
```
slavek-kucera, line 546: when does this happen?

SWETAS04: There is a test "parse_bad_model" where it occurs

parser_library/src/diagnostic.cpp
```diff
@@ -1857,6 +1864,13 @@ diagnostic_op diagnostic_op::error_ME002(const range& range)
     return diagnostic_op(diagnostic_severity::error, "ME002", "multiplication or division of address", range);
 }
 
+diagnostic_op diagnostic_op::error_ME003(const range& range)
+{
+    return diagnostic_op(diagnostic_severity::error,
+        "ME003",
+        "Relative Immediate Operand value must be signed number of halfwords.",
```
michalbali256, line 1871: This message implies to me, that the user should write a number as the argument.
```suggestion
        "Relative immediate operand must evaluate into a number of halfwords (must result in even number of bytes)",
```

slavek-kucera (approved):
👍

parser_library/src/semantics/operand_impls.cpp
```diff
@@ -92,18 +91,24 @@ std::unique_ptr<checking::operand> make_check_operand(expressions::mach_evaluate
...
-        else
-        {
-            return std::make_unique<checking::address_operand>(
-                checking::address_state::UNRES, 0, 0, 0, checking::operand_state::ONE_OP);
-        }
+        return std::make_unique<checking::address_operand>(
+            checking::address_state::UNRES, 0, 0, 0, checking::operand_state::ONE_OP);
+    }
+}
+std::unique_ptr<checking::operand> check_operand(
```
michalbali256, line 98: could you rename this method? Now, it is not clear what is the difference between methods `check_operand` and `make_check_operand`

parser_library/src/diagnostic.cpp
slavek-kucera: maybe just `Relative immediate operand must evaluate into an even offset`?

michalbali256 (approved):

PR #129, opened 2021-05-10
perf: Dependency files caching

michalbali256:
Cache dependencies (macro and COPY files), so they are not parsed multiple times, if the user has not made any changes to them.


parser_library/src/context/copy_member.h
```diff
@@ -58,7 +42,25 @@ struct copy_member
...
     }
 
-    copy_member_invocation enter() { return copy_member_invocation(name, cached_definition, definition_location); }
+    // copy_member_invocation enter() { return copy_member_invocation(name, cached_definition, definition_location); }
+};
+
+// structure represents invocation of COPY member in HLASM macro library
+struct copy_member_invocation
+{
+    const id_index name;
```
slavek-kucera, line 51: what was the reason to make these const?

parser_library/src/workspaces/processor_file_impl.cpp
```diff
@@ -42,16 +47,20 @@ bool processor_file_impl::is_once_only() const { return false; }
 
 parse_result processor_file_impl::parse(parse_lib_provider& lib_provider)
 {
-    analyzer_ = std::make_unique<analyzer>(get_text(), get_file_name(), lib_provider, get_lsp_editing());
+    if (opencode_analyzer_)
+        opencode_analyzer_ = std::make_unique<analyzer>(
+            get_text(), get_file_name(), lib_provider, get_lsp_editing(), opencode_analyzer_->hlasm_ctx().move_ids());
```
slavek-kucera, line 52: why not just?
```suggestion
            get_text(), get_file_name(), lib_provider, get_lsp_editing(), std::move(opencode_analyzer_->hlasm_ctx().ids()));
```

parser_library/src/workspaces/processor_file_impl.h
```diff
@@ -40,26 +43,27 @@ class processor_file_impl : public virtual file_impl, public virtual processor_f
...
     const lsp::feature_provider& get_lsp_feature_provider() override;
     const std::set<std::string>& files_to_close() override;
     const performance_metrics& get_metrics() override;
 
+    void erase_cache_of_opencode(const std::string& opencode_file_name) override;
+
 private:
-    std::unique_ptr<analyzer> analyzer_;
+    std::unique_ptr<analyzer> opencode_analyzer_;
+    const analyzer* last_analyzer_;
```
slavek-kucera, line 57: so the purpose of this is to retrieve results from the last processing?

parser_library/src/workspaces/processor_file_impl.h
```diff
@@ -40,26 +43,27 @@ class processor_file_impl : public virtual file_impl, public virtual processor_f
...
 
     bool parse_inner(analyzer&);
 
-    bool parse_info_updated_ = false;
     std::atomic<bool>* cancel_;
 
     std::set<std::string> dependencies_;
     std::set<std::string> files_to_close_;
```
slavek-kucera, line 64: can this be impacted by different environments during processing? In other word will this "match" the content in the cache?

parser_library/src/lsp/lsp_context.cpp
```diff
@@ -106,6 +106,18 @@ void lsp_context::add_opencode(opencode_info_ptr opencode_i, text_data_ref_t tex
     distribute_file_occurences(opencode_->file_occurences);
 }
 
+macro_info_ptr lsp_context::get_macro_info(context::id_index macro_name) const
+{
+    if (!opencode_)
```
slavek-kucera, line 111: what are the conditions in which the `opencode_` is empty in this function?

parser_library/src/workspaces/macro_cache.cpp
```diff
@@ -0,0 +1,202 @@
...
+        wn.SETB,
+        wn.SETC,
+        ctx.ids().add("AIF"),
+        ctx.ids().add("MEND"),
+        ctx.ids().add("MACRO"),
+        ctx.ids().add("MEXIT"),
+        ctx.ids().add("AIF"),
+        ctx.ids().add("AREAD"),
+        ctx.ids().add("ACTR"),
+        ctx.ids().add("AGO") };
```
slavek-kucera, line 55: Whould it make sense to add these to the well_known set?

parser_library/src/workspaces/macro_cache.cpp
```diff
@@ -0,0 +1,202 @@
...
+            result.push_back(
+                { from, opcode.opcode, std::holds_alternative<context::macro_def_ptr>(opcode.opcode_detail) });
+
+        if (std::find(cached_instr.begin(), cached_instr.end(), from) != cached_instr.end())
+            result.push_back(
+                { from, opcode.opcode, std::holds_alternative<context::macro_def_ptr>(opcode.opcode_detail) });
+    }
+
+    // Also macros with the same name as CA instructions may alias the instructions
+    for (const auto& [id, macro] : ctx.macros())
```
slavek-kucera, line 72: Are these inline defined macro only?

parser_library/src/workspaces/macro_cache.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+bool macro_cache::load_from_cache(const macro_cache_key& key, const analyzing_context& ctx)
+{
+    if (auto cached_analyzer = find_cached_analyzer(key))
+    {
+        if (key.data.proc_kind == processing::processing_kind::MACRO)
+        {
+            lsp::macro_info_ptr info = cached_analyzer->context().lsp_ctx->get_macro_info(key.data.library_member);
+            if (!info)
+                return true; // The file for which the analyzer is cached does not contain definition of macro
```
slavek-kucera, line 133: so is this a case of an invalid file in a macro library?

parser_library/src/workspaces/macro_cache.cpp
```diff
@@ -0,0 +1,202 @@
...
+    auto it = cache_.find(key);
+    if (it == cache_.end())
+        return nullptr;
+
+    const auto& cached_data = it->second;
+
+    for (const auto& [fname, cached_version] : cached_data.stamps)
+    {
+        auto file = file_mngr_->find(fname);
+        if (!file)
```
slavek-kucera, line 115: could we keep like weak_ptr to the file instead of file_manager ref and the name?

parser_library/src/workspaces/macro_cache.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+version_stamp macro_cache::get_copy_member_versions(context::macro_def_ptr macro) const
+{
+    version_stamp result;
+
+    for (const auto& copy_ptr : macro->used_copy_members)
+    {
+        auto file = file_mngr_->find(copy_ptr->definition_location.file);
+        if (!file)
+            throw std::runtime_error("Dependencies of a macro must be open right after parsing the macro.");
```
slavek-kucera, line 165: how does this happen?

parser_library/src/workspaces/macro_cache.cpp
```diff
@@ -0,0 +1,202 @@
...
+    else // Copy members do not have additional dependencies
+        cache_data.stamps.clear();
+
+    cache_data.stamps.try_emplace(macro_file_->get_file_name(), macro_file_->get_version());
+    cache_data.cached_analyzer = std::move(analyzer);
+}
+
+void macro_cache::erase_cache_of_opencode(const std::string opencode_file_name)
+{
+    auto it = cache_.begin();
```
slavek-kucera, line 192: any concurrency concerns?

parser_library/src/workspaces/macro_cache.h
```diff
@@ -0,0 +1,102 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Broadcom, Inc. - initial API and implementation
+ */
+
+#include "analyzer.h"
+
+#ifndef HLASMPLUGIN_PARSERLIBRARY_MACRO_CACHE_H
+#    define HLASMPLUGIN_PARSERLIBRARY_MACRO_CACHE_H
```
slavek-kucera, line 18:
```suggestion
#define HLASMPLUGIN_PARSERLIBRARY_MACRO_CACHE_H
```

parser_library/test/workspace/macro_cache_test.cpp
```diff
@@ -0,0 +1,337 @@
...
+    auto macro_diag = std::find_if(opencode->diags().begin(), opencode->diags().end(), [&](const diagnostic_s& d) {
+        return d.file_name == macro_file_name;
+    });
+    EXPECT_NE(macro_diag, opencode->diags().end());
+
+    auto opencode_diag = std::find_if(opencode->diags().begin(), opencode->diags().end(), [&](const diagnostic_s& d) {
+        return d.file_name == opencode_file_name;
+    });
+    EXPECT_NE(opencode_diag, opencode->diags().end());
+}
```
slavek-kucera, line 337: maybe a test case where inline macro depends on external COPY?

parser_library/src/context/copy_member.h
michalbali256: I think the only reason was that it is a structure and it does not make sense to change these members. Getter functions would be certainly better.

parser_library/src/lsp/lsp_context.cpp
michalbali256: It can be empty when parsing is interrupted by cancellation token.
And also when there is a corrupted macro files, so there is no macro definition in it and open code calls the macro twice, the first time we cache the resulting analyzer with no macro definition. The second call reads the cache and tries to get the macro info through this function, but the opencode is not there yet.

Although the second case seems strange, I didn't fix, since it won't happen very often and we issue a diagnostic that the macro file should define a macro.

parser_library/src/workspaces/macro_cache.cpp
michalbali256: No, the structure contains all macro definitions that were parsed up until the particular moment if parsing.

parser_library/src/workspaces/macro_cache.cpp
michalbali256: Probably not, since file_manager creates new shared pointers to processor_files in the rare cases when the files are accessed concurrently. We need to access the newest version here.

parser_library/src/workspaces/macro_cache.cpp
michalbali256: yes, exactly

parser_library/src/workspaces/macro_cache.cpp
michalbali256: I don't see that it could happen.. Should I remove the condition?

parser_library/src/workspaces/macro_cache.cpp
michalbali256: the function is currently called from the workspace, which does not have any concurrent usage, as far as I know

parser_library/src/workspaces/processor_file_impl.h
michalbali256: The `dependencies_` and `files_to_close_` members are only used by files that are parsed as open codes, so they should not be affected by macro cache.

parser_library/src/context/id_storage.cpp
```diff
@@ -69,6 +69,11 @@ hlasm_plugin::parser_library::context::id_storage::well_known_strings::well_know
     , LCLC(&*ptr.emplace("LCLC").first)
     , MACRO(&*ptr.emplace("MACRO").first)
     , MEND(&*ptr.emplace("MEND").first)
+    , MEXIT(&*ptr.emplace("MEXIT").first)
     , ASPACE(&*ptr.emplace("ASPACE").first)
+    , AIF(&*ptr.emplace("AIF").first)
+    , ACTR(&*ptr.emplace("ACTR").first)
+    , AREAD(&*ptr.emplace("AREAD").first)
+    , AGO(&*ptr.emplace("AGO").first)
```
slavek-kucera, line 77: to match the actual initialization order.
```suggestion
    , AGO(&*ptr.emplace("AGO").first)
    , ACTR(&*ptr.emplace("ACTR").first)
    , AREAD(&*ptr.emplace("AREAD").first)
```

parser_library/src/workspaces/macro_cache.cpp
slavek-kucera: I am again thinking about a race condition or something like that...

parser_library/src/workspaces/macro_cache.h
```diff
@@ -0,0 +1,102 @@
...
+// Pair of version stamp and analyzer that parsed the version of file(s)
+struct macro_cache_data
+{
+    // Versions of respective macro (copy member) with all dependencies (COPY instruction is evaluated during macro
+    // definition and the statements are part of macro definition)
+    version_stamp stamps;
+    std::unique_ptr<analyzer> cached_analyzer;
+};
+
+class macro_cache : public diagnosable_impl
```
slavek-kucera, line 78: can be final?
```suggestion
class macro_cache final : public diagnosable_impl
```

PR #134, opened 2021-06-03
refactor: Split complex classes

slavek-kucera:
refactor: parser_impl split
refactor: EOLLN mechanism replaced by true per logical line parsing
fix: apostrophe in strings may be incorrectly reparsed
fix: reparse stops at NUL characters
fix: ignore diags and highlighting info in the lookahead mode
fix: correctly diagnose invalid line continuations
fix: crash on an invalid instruction in the lookahead mode
fix: follow argument restrictions of isspace, isaplha etc. 
feat: Lines processed by AREAD are highlighted as such
test: hover over non-existent macro
refactor: unify analyzer options
fix: handle empty variables and concatenations in the label


parser_library/include/range.h
```diff
@@ -40,6 +40,21 @@ struct PARSER_LIBRARY_EXPORT position
     bool operator!=(const position& oth) const { return !(*this == oth); }
     position_t line;
     position_t column;
+
+    static inline position min(const position& lhs, const position& rhs)
```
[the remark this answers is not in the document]
slavek-kucera: Given the size of the structure, I just prefer returning a value and it also protects from accidentally capturing dead rvalue with const auto&.

parser_library/src/parsing/parser_error_listener_ctx.h
```diff
@@ -31,6 +31,13 @@ class parser_error_listener_ctx : public diagnosable_ctx, public parser_error_li
         : diagnosable_ctx(hlasm_ctx)
         , parser_error_listener_substitution(add_diag_with_ctx, substituted, std::move(provider))
     {}
+    parser_error_listener_ctx(context::hlasm_context& hlasm_ctx,
```
michalbali256, line 34: Why would I want to use this overload instead of parser_error_listener_substitution? It completely bypasses the added value of context.

parser_library/src/processing/opencode_provider.cpp
```diff
@@ -0,0 +1,563 @@
...
+{
+    auto line_no = m_current_logical_line_source.begin_line;
+    for (const auto& s : m_current_logical_line.segments)
+    {
+        if (s.continuation_error)
+        {
+            parsing::parser_error_listener_ctx listener(*m_ctx->hlasm_ctx, nullptr);
+            listener.add_diagnostic(
+                diagnostic_op::error_E001(range { { line_no, 0 }, { line_no, s.code_offset_utf16 } }));
+            collect_diags_from_child(listener);
```
michalbali256, line 195: Why not make opencode_provider inherit from diagnosable_ctx? Then these lines would be just:
```suggestion
            add_diagnostic(
                diagnostic_op::error_E001(range { { line_no, 0 }, { line_no, s.code_offset_utf16 } }));
```

parser_library/src/processing/opencode_provider.h
```diff
@@ -26,20 +45,122 @@ enum class ainsert_destination
...
+    void rewind_input(context::source_position pos);
+    std::string aread();
+    void ainsert(const std::string& rec, ainsert_destination dest);
+
+    opencode_provider(std::string_view text,
+        analyzing_context& ctx,
+        workspaces::parse_lib_provider& lib_provider,
+        processing::processing_state_listener& state_listener,
+        semantics::source_info_processor& src_proc,
+        const std::string& filename,
```
michalbali256, line 127: I am not 100% about this, but context has method `opencode_file_name`, probably could replace this parameter.

slavek-kucera: there appear to be cases when we are passing both the context and the filename as parameters and at the moment I am not 100% convinced that they always match.

PR #138, opened 2021-06-21
fix: AREAD/AINSERT support in macros called from copybooks

slavek-kucera:
fix: Address edge cases when AREAD and AINSERT are called from source code brought by COPY instruction.
feat: Add support for macro library to fuzzer.

### DIFF
--- a/parser_library/src/lsp/text_data_ref_t.cpp
+++ b/parser_library/src/lsp/text_data_ref_t.cpp
@@ -41,8 +41,6 @@ std::string_view text_data_ref_t::get_line_beginning(position pos) const
     return { &text->at(line_indices[pos.line]), line_len };
 }
 
-
-
 char text_data_ref_t::get_character_before(position pos) const
 {
     if (pos.column == 0)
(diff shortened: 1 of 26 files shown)


parser_library/test/processing/aread_test.cpp
```diff
@@ -359,3 +359,385 @@ Line5
...
+public:
+    parse_result parse_library(const std::string& library, analyzing_context ctx, library_data data) override
+    {
+        auto it = m_files.find(library);
+        if (it == m_files.end())
+            return false;
+
+        auto a = std::make_unique<analyzer>(it->second, analyzer_options { library, this, std::move(ctx), data });
+        a->analyze();
+        a->collect_diags();
```
michalbali256, line 377:
```suggestion
```

slavek-kucera: We end up doing this step during normal processing, so while it is redundant here, I'd prefer keeping it.

parser_library/src/lsp/text_data_ref_t.cpp
```diff
@@ -60,6 +58,13 @@ std::string_view text_data_ref_t::get_range_content(range r) const
     return std::string_view(&text->at(start_i), end_i - start_i);
 }
 
+std::string_view text_data_ref_t::get_lines_beginning(position pos) const
```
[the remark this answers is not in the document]
slavek-kucera: actually, I wanted `get_lines_beginning_at` and `get_line_beginning_at` to have the same signature and did not want to spend time changing the later. So, I'd prefer to keep the position argument for now.

michalbali256 (approved):

PR #141, opened 2021-06-30
refactor: Parser reports all its diagnostics using single channel

michalbali256:
Previously, some diagnostics were reported using `diagnosable` pattern and some using `parser_error_listener`. This PR introduces interface diagnostic consumer, so the owner of parser decides where the diagnostics will end up.


parser_library/src/diagnostic_consumer.h
```diff
@@ -0,0 +1,70 @@
...
+
+template<typename T>
+class diagnostic_consumer_container : public diagnostic_consumer<T>
+{
+public:
+    mutable std::vector<T> diags;
+    void add_diagnostic(T diagnostic) const override { diags.push_back(std::move(diagnostic)); };
+};
+
+using diagnostic_op_consumer_container = diagnostic_consumer_container<diagnostic_op>;
```
slavek-kucera, line 66: is it used only in tests?

parser_library/src/expressions/evaluation_context.h
```diff
@@ -34,7 +34,12 @@ struct evaluation_context : diagnosable_ctx
         , lib_provider(lib_provider)
     {}
 
-    evaluation_context(const evaluation_context&) = delete;
+    evaluation_context(const evaluation_context& oth)
+        : diagnosable_ctx(oth)
+        , ctx(oth.ctx)
+        , hlasm_ctx(*ctx.hlasm_ctx)
+        , lib_provider(oth.lib_provider)
+    {}
```
slavek-kucera, line 42: isn't this the `default` implementation and why do we suddenly need it?

parser_library/src/processing/statement_fields_parser.cpp
```diff
@@ -55,20 +52,36 @@ const parsing::parser_holder& statement_fields_parser::prepare_parser(const std:
...
+    diagnostic_op_consumer* diagnoser;
+    std::string* field;
+    void add_diagnostic(diagnostic_op diag) const override
+    {
+        if (field)
+            diag.message = "While substituting to '" + *field + "' => " + diag.message;
+        diagnoser->add_diagnostic(std::move(diag));
+    }
+};
+
```
slavek-kucera, line 67: is it used/how does it differ from diagnostic_op_consumer_transform?
```suggestion

```

slavek-kucera (changes requested):

parser_library/src/processing/statement_fields_parser.cpp
```diff
@@ -55,20 +52,36 @@ const parsing::parser_holder& statement_fields_parser::prepare_parser(const std:
...
-    parsing::parser_error_listener_ctx listener(
-        *m_hlasm_ctx, after_substitution ? &field : nullptr, add_diag, field_range);
-
     const auto original_range = field_range.original_range;
 
-    const auto& h = prepare_parser(field, after_substitution, std::move(field_range), status, listener);
+
+    diagnostic_op_consumer_transform add_diag_subst([&](diagnostic_op diag) {
+        if (after_substitution)
+            diag.message = "While substituting to '" + field + "' => " + diag.message;
```
slavek-kucera, line 81: Won't `While substituting to` be followed by the result of substitution? Making the phrasing a bit confusing?

parser_library/src/diagnostic_consumer.h
michalbali256: currently yes, but it may be useful in production in the future so I left it here.

slavek-kucera (approved):

parser_library/src/processing/statement_fields_parser.cpp
```diff
@@ -59,16 +56,19 @@ std::pair<semantics::operands_si, semantics::remarks_si> statement_fields_parser
...
-
     const auto original_range = field_range.original_range;
 
-    const auto& h = prepare_parser(field, after_substitution, std::move(field_range), status, listener);
+
+    diagnostic_op_consumer_transform add_diag_subst([&](diagnostic_op diag) {
+        if (after_substitution)
+            diag.message = "While evaluating the result of substitution '" + field + "' => " + diag.message;
+        add_diag.add_diagnostic(std::move(diag));
+    });
```
slavek-kucera, line 70:
```suggestion
    diagnostic_op_consumer_transform add_diag_subst([&field, after_substitution](diagnostic_op diag) {
        if (after_substitution)
            diag.message = "While evaluating the result of substitution '" + field + "' => " + std::move(diag.message);
        add_diag.add_diagnostic(std::move(diag));
    });
```
and re-run the benchmark?

slavek-kucera (approved):

PR #147, opened 2021-07-15
feat : Allow wildcards to be specified in the proc_grps settings for macro resolution across multiple folders

SWETAS04:


parser_library/src/workspaces/wildcard.cpp
```diff
@@ -38,5 +42,44 @@ std::regex wildcard2regex(const std::string& wildcard)
...
+    auto prefix_path = lib_path.substr(0, pos);
+    pos = pos + 1;
+    for (const auto& p : std::filesystem::recursive_directory_iterator(prefix_path))
+    {
+        std::string suffix_path = lib_path.substr(pos);
+        suffix_path.pop_back();
+        suffix_path = std::regex_replace(suffix_path, std::regex("\\*"), ".*");
+
+
+        std::string directory_path = p.path().lexically_normal().generic_string();
```
slavek-kucera, line 58: I am not sure this will work correctly in the WASM variant. See utils/include/utils/* and utils/src/* for more details...

slavek-kucera (commented):
For this feature we will need a test that runs in the vscode - to verify of behavior both native and WASM targets.

Does the implementation handle symbolic links?
Is there a limit for the depth of the recursion in case there is a cycle on the filesystem created by e.g. symlinks?

PR #148, opened 2021-07-15
fix: Inline macros overwriting external definition stored in macro cache

michalbali256:
fix: inline macros overwriting external definition stored in macro cache
refactor: evaluation context does not need lsp context, only hlasm context
fix: crash when `processor_file` is queried for information prior to first parse

### DIFF
--- a/parser_library/include/protocol.h
+++ b/parser_library/include/protocol.h
@@ -223,13 +223,14 @@ struct PARSER_LIBRARY_EXPORT performance_metrics
     size_t files = 0;
 };
 
-bool inline operator==(const performance_metrics& lhs, const performance_metrics& oth)
+bool inline operator==(const performance_metrics& lhs, const performance_metrics& rhs)
 {
-    return lhs.lines == oth.lines && lhs.macro_def_statements == oth.macro_def_statements
-        && lhs.open_code_statements == oth.open_code_statements && lhs.copy_def_statements == oth.copy_def_statements
-        && lhs.copy_statements == oth.copy_statements && lhs.reparsed_statements == oth.reparsed_statements
-        && lhs.lookahead_statements == oth.lookahead_statements && lhs.continued_statements == oth.continued_statements
-        && lhs.non_continued_statements == oth.non_continued_statements && lhs.files == oth.files;
+    return lhs.lines == rhs.lines && lhs.macro_def_statements == rhs.macro_def_statements
+        && lhs.macro_statements == rhs.macro_statements && lhs.open_code_statements == rhs.open_code_statements
+        && lhs.copy_def_statements == rhs.copy_def_statements && lhs.copy_statements == rhs.copy_statements
+        && lhs.reparsed_statements == rhs.reparsed_statements && lhs.lookahead_statements == rhs.lookahead_statements
+        && lhs.continued_statements == rhs.continued_statements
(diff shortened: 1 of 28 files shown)


parser_library/include/protocol.h
```diff
@@ -223,6 +223,15 @@ struct PARSER_LIBRARY_EXPORT performance_metrics
     size_t files = 0;
 };
 
+bool inline operator==(const performance_metrics& lhs, const performance_metrics& oth)
+{
+    return lhs.lines == oth.lines && lhs.macro_def_statements == oth.macro_def_statements
```
slavek-kucera, line 228: macro_statements missing?

parser_library/src/workspaces/processor_file_impl.cpp
```diff
@@ -140,17 +140,57 @@ const std::set<std::string>& processor_file_impl::dependencies() { return depend
...
+    {
+        return lsp::hover_result();
+    }
+    virtual lsp::completion_list_s completion(const std::string& document_uri,
+        position pos,
+        char trigger_char,
+        completion_trigger_kind trigger_kind) const override
+    {
+        return {};
+    }
```
slavek-kucera, line 172: drop the virtual

slavek-kucera: also I wonder whether we should have the capability of returning "we don't know yet".

parser_library/src/lsp/lsp_context.cpp
```diff
@@ -108,13 +112,9 @@ void lsp_context::add_opencode(opencode_info_ptr opencode_i, text_data_ref_t tex
...
 {
-    // Opencode may be empty, if parsing was interrupted by cancellation token, or when this function is called in the
-    // middle of processing
-    if (!opencode_)
-        return nullptr;
     // This function does not respect OPSYN, so we do not use hlasm_context::get_macro_definition
-    auto it = opencode_->hlasm_ctx.macros().find(macro_name);
-    if (it == opencode_->hlasm_ctx.macros().end())
+    auto it = hlasm_ctx_->macros().find(macro_name);
+    if (it == hlasm_ctx_->macros().end())
```
slavek-kucera, line 117: could this be responsible for the perf. drop?

parser_library/src/analyzer.cpp
```diff
@@ -27,11 +27,12 @@ analyzing_context& analyzer_options::get_context()
...
+        auto h_ctx = std::make_shared<context::hlasm_context>(file_name,
+            std::move(std::get<asm_option>(ctx_source)),
+            ids_init.has_value() ? std::move(*ids_init) : id_storage {});
         ctx_source = analyzing_context {
-            std::make_unique<context::hlasm_context>(file_name,
-                std::move(std::get<asm_option>(ctx_source)),
-                ids_init.has_value() ? std::move(*ids_init) : id_storage {}),
-            std::make_unique<lsp::lsp_context>(),
+            h_ctx,
+            std::make_unique<lsp::lsp_context>(h_ctx),
```
slavek-kucera, line 35:
```suggestion
            std::make_shared<lsp::lsp_context>(h_ctx),
```

slavek-kucera (approved):

slavek-kucera (approved):

PR #170, opened 2021-08-20
fix: Diagnostics lost during JSON serialization

slavek-kucera:


parser_library/src/processing/statement_fields_parser.cpp
```diff
@@ -64,7 +112,7 @@ std::pair<semantics::operands_si, semantics::remarks_si> statement_fields_parser
 
     diagnostic_consumer_transform add_diag_subst([&field, &add_diag, after_substitution](diagnostic_op diag) {
         if (after_substitution)
-            diag.message = "While evaluating the result of substitution '" + field + "' => " + std::move(diag.message);
+            diag.message = decorate_message(field, diag.message);
```
michalbali256, line 115: Do we want to run the sanitizer in the language_server to sanitize all diagnostics? I think there are some more diagnostics that show names of symbols... Although those are probably guaranteed to be 'nice' strings since they were lexed as ordinary symbols, it may be more futureproof solution, if we decide to add different diagnostics in the future.

michalbali256: Also, there are more occasions where we copy the source code into json messages, see `lsp_context::get_macro_documentation`, and maybe some other places in `lsp_context`

slavek-kucera: As far as I can tell, the only instance when we may end up including an invalid utf-8 sequences into diagnostics is when the code is generated (e.g. by substitutions).
In all other cases the text should be already sanitized or even parsed.

michalbali256: Yes you are right, I forgot that we sanitize the source prior to parsing it.

How about values of variables that we send via DAP?

slavek-kucera: I'll look into that one... 👍

slavek-kucera: yes, so dap variables are affected as well. But we could address that together with other dap issues that need to be solved.

michalbali256 (approved):

PR #171, opened 2021-08-27
feat: Support START and MHELP instructions

slavek-kucera:


parser_library/test/processing/start_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+    a.collect_diags();
+    EXPECT_TRUE(matches_message_codes(a.diags(), { "E073" }));
+}
+
+TEST(START, dsect_not_csect)
+{
+    std::string input(R"(
+D DSECT
+E EQU 0
+S CSECT
```
michalbali256, line 103:
```suggestion
S START
```
Otherwise, I don't see how this tests the START instruction

parser_library/src/processing/instruction_sets/asm_processor.cpp
```diff
@@ -765,4 +766,52 @@ void asm_processor::process_CNOP(rebuilt_statement stmt)
...
+    auto sym_loc = hlasm_ctx.processing_stack().back().proc_location;
+    sym_loc.pos.column = 0;
+    hlasm_ctx.ord_ctx.set_section(sect_name, context::section_kind::EXECUTABLE, std::move(sym_loc));
+
+    const auto& ops = stmt.operands_ref().value;
+    if (ops.size() != 1)
+        return;
+
+    auto initial_offset = try_get_abs_value(ops.front().get());
+    if (!initial_offset.has_value())
```
michalbali256, line 800: Emit a diagnostic here that says "operand has to be absolute expression with all symbols previously defined"? It is a special requirement that will probably never be checked in the checker.

michalbali256 (approved):

PR #185, opened 2021-09-22
fix: Operands of dynamically generated statements may be incorrectly parsed

slavek-kucera:


michalbali256 (approved):

parser_library/src/context/statement_cache.h
```diff
@@ -37,7 +37,8 @@ class statement_cache
     };
     // pair of processing format and reparsed statement
     // processing format serves as an identifier of reparsing kind
-    using cache_t = std::pair<processing::processing_form, cached_statement_t>;
+    using cache_t =
+        std::pair<std::pair<processing::processing_form, processing::operand_occurence>, cached_statement_t>;
```
slavek-kucera, line 41: the whole processing_status might need to be used as a key...

michalbali256: Now, that we are asking for specific instructions during parsing (ALIAS) you are probably right...
processing_kind probably doesn't matter, but doesn't hurt either.

michalbali256: Or maybe just op_code is enough?

slavek-kucera: parser_impl::resolve_expression is impacted by as well...

parser_library/test/context/macro_test.cpp
```diff
@@ -859,3 +859,30 @@ TEST(macro, missing_mend)
...
+         AIF   ('&THIRD' EQ '').SKIP
+&OPERAND SETC  '&FIRST.,&SECOND.,&THIRD.'
+.SKIP    ANOP
+         &OP   &OPERAND
+         MEND
+
+         MAC SAM64,this_is_comment
+         MAC LA,0,0
+         MAC ARK,0,0,0
+
```
michalbali256, line 881:
```suggestion
         MAC ALIAS,C'somealias'
ALIAS OPSYN SAM64
         MAC ALIAS,this_is_comment
```

I think alias should be explicitly tested as well, since it is handled in a special way.

Also adding opsyn to the mix might be a good idea.

michalbali256 (approved):

PR #187, opened 2021-09-24
feat: Telemetry reporting

michalbali256:


clients/vscode-hlasmplugin/src/extension.ts
```diff
@@ -76,10 +105,13 @@ export async function activate(context: vscode.ExtensionContext) {
 
     if (serverVariant === 'native')
         startCheckingNativeClient(hlasmpluginClient);
-
+
```
slavek-kucera, line 108:
```suggestion

```

language_server/src/dap/dap_server.cpp
```diff
@@ -71,6 +71,7 @@ void server::message_received(const json& message)
         if (message.at("type") != "request")
         {
             LOG_WARNING(std::string("Invalid message receive: ") + message.dump());
+            telemetry_error("lsp_server/invalid_message");
```
slavek-kucera, line 74:
```suggestion
            telemetry_error("dap_server/invalid_message");
```

language_server/src/server.h
```diff
@@ -64,11 +67,25 @@ class server : public response_provider
...
     // Calls a method that is registered in methods_ with the specified name with arguments and id of request.
     void call_method(const std::string& method, const json& id, const json& args);
+
+    struct telemetry_metrics_info
+    {
+        json properties;
+        json metrics;
+    };
+
+    telemetry_metrics_info virtual get_telemetry_details();
```
slavek-kucera, line 83: This seems a bit strange... how is the method supposed to know to which request the details are being provided?

language_server/src/lsp/lsp_server.h
```diff
@@ -41,6 +43,11 @@ class server final : public hlasm_plugin::language_server::server,
     // Parses LSP (JSON RPC) message and calls corresponding method.
     void message_received(const json& message) override;
 
+    // Inherited via json_sink
+    // Takes json with telemetry info and sends it through LSP
+    void write(const nlohmann::json&) override;
+    void write(nlohmann::json&&) override;
+
```
slavek-kucera, line 50: wouldn't it be better to have a dedicated interface that takes a telemetry_info type?

slavek-kucera (changes requested):
Some of the code smells should probably be addressed.

clients/vscode-hlasmplugin/src/extension.ts
```diff
@@ -23,22 +23,46 @@ import { CustomEditorCommands } from './customEditorCommands';
...
  */
 export async function activate(context: vscode.ExtensionContext) {
+    const serverVariant = getConfig<ServerVariant>('serverVariant', 'native');
+    
+    var telemetry = new Telemetry();
+    context.subscriptions.push(telemetry);
+    
+    // setTimeout is needed, because telemetry initialization is asynchronous
+    // and AFAIK no event in the API is exposed to send the activation telemetry event
+    setTimeout(() => {telemetry.reportEvent("hlasm.activated", {server_variant:serverVariant.toString()});}, 1000);
```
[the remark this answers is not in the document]
michalbali256: I don't think so. I wanted to have an event that would trigger right after activation and would not depend on e.g. ability to run native executables.

BTW creators of vscode-extension-telemetry acknowledged that this is a shortfall of the plugin and plan to either introduce an event or buffer where such event would be stored before they can be sent. https://github.com/microsoft/vscode-extension-telemetry/issues/76

clients/vscode-hlasmplugin/src/extension.ts
```diff
@@ -50,16 +74,21 @@ export async function activate(context: vscode.ExtensionContext) {
             code2Protocol: (value: vscode.Uri) => value.toString(),
             protocol2Code: (value: string) =>
                 vscode.Uri.file((vscode.Uri.parse(value).fsPath))
-        }
+        },
+        errorHandler: clientErrorHandler
```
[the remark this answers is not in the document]
michalbali256: Yes it does

language_server/src/server.h
michalbali256: I agree it is a bit strange... the details are provided for the last request.

It fetches the details for requests with `telemetry_log_level::LOG_WITH_METRICS`, which is actually only one request (didOpen).

I only wanted the `lsp_server` to have the metadata consumer, but at the same time I wanted to log the requests in the `server::call_method`, method, which is common for lsp and dap server. Thats why I ended up with this virtual method.

slavek-kucera (approved):

PR #194, opened 2021-10-12
Introduce a limit to outline tree size

slavek-kucera:


language_server/src/lsp/feature_language_features.cpp
```diff
@@ -320,7 +320,8 @@ void feature_language_features::document_symbol(const json& id, const json& para
 {
     auto document_uri = params["textDocument"]["uri"].get<std::string>();
 
-    auto symbol_list = ws_mngr_.document_symbol(uri_to_path(document_uri).c_str());
+    const auto limit = 5000LL;
+    auto symbol_list = ws_mngr_.document_symbol(uri_to_path(document_uri).c_str(), limit);
```
michalbali256, line 324: Would it be better to make the limit configurable by adding it to the user configuration and then using the configuration changed notification (which is already implemented). Might be even easier, since no edits to workspace_manager would be required.

michalbali256: Maybe I would prefer that workspace would not implement the interface and put the limit into lib_config. But then lib_config wouldn't represent only user-configurable options.

I guess I am fine with both versions, then.

PR #208, opened 2021-12-10
feat: Location counter length attribute support

### LINKED ISSUE #71: Length attribute of the location counter reference
Add missing functionality that allows you to use the length attribute of the location counter reference (L'*).

image

slavek-kucera:
(closes #71)

### DIFF
--- a/parser_library/src/parsing/parser_impl.cpp
+++ b/parser_library/src/parsing/parser_impl.cpp
@@ -126,6 +126,11 @@ size_t parser_impl::get_loctr_len() const
     return processing::processing_status_cache_key::generate_loctr_len(opcode.value);
 }
 
+bool parser_impl::loctr_len_allowed(const std::string& attr) const
+{
+    return (attr == "L" || attr == "l") && proc_status.has_value();
+}
+
 void parser_impl::resolve_expression(expressions::ca_expr_ptr& expr, context::SET_t_enum type) const
 {
     expr->resolve_expression_tree(type);
--- a/parser_library/src/parsing/parser_impl.cpp
+++ b/parser_library/src/parsing/parser_impl.cpp
@@ -23,6 +23,7 @@
 #include "hlasmparser.h"
 #include "lexing/token_stream.h"
 #include "processing/context_manager.h"
+#include "processing/op_code.h"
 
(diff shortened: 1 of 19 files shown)


parser_library/test/processing/mach_instr_test.cpp
```diff
@@ -254,3 +254,30 @@ TEST(mach_instr_processing, rel_addr_bitmask)
...
+             { "CLIJ", 6 },
+             { "BR", 2 },
+             { "DC", 1 },
+         })
+    {
+        EXPECT_EQ(processing::processing_status_cache_key::generate_loctr_len(&instr), expected) << instr;
+    }
+}
+
+TEST(mach_instr_processing, loctr_len_reference)
```
michalbali256, line 272: Could you think of a test verifying the `L'*` is resolved after resolving opsyn?

parser_library/src/parsing/parser_impl.cpp
```diff
@@ -119,6 +120,12 @@ context::id_index parser_impl::parse_identifier(std::string value, range id_rang
     return hlasm_ctx->ids().add(std::move(value));
 }
 
+size_t parser_impl::get_loctr_len() const
+{
+    auto [_, opcode] = *proc_status;
+    return processing::processing_status_cache_key::generate_loctr_len(opcode.value);
```
michalbali256, line 126: Is the processing status/opcode valid when parsing something like
```
&VARP(L'=CL(L'*)'1') SETA 1
```

How does the assembler handle `L'*` in labels?

slavek-kucera: It currently does not, but IBM has released an APAR PH42377 that enhances parsing of literal attributes, so this will have to be verified again once we get it on our systems.
I'll add a test and fix the problem.

parser_library/test/processing/mach_instr_test.cpp
slavek-kucera: I think the test above does exactly that and shows a bug. We'll have to wait for the APAR mentioned to replicate the correct new behavior.

michalbali256 (approved):

PR #223, opened 2022-01-27
fix: Data definition grammar is too greedy

slavek-kucera:


parser_library/src/diagnostic.cpp
```diff
@@ -1363,15 +1363,16 @@ diagnostic_op diagnostic_op::error_D003(const range& range)
         diagnostic_severity::error, "D003", "Expected an integer or an expression after modifier", range);
 }
 
-diagnostic_op diagnostic_op::error_D004(const range& range)
-{
-    return diagnostic_op(diagnostic_severity::error, "D004", "Wrong order of modifiers", range);
-}
+// diagnostic_op diagnostic_op::error_D004(const range& range)
+// {
```
SWETAS04, line 1367: Are we going to keep this commented code ?

slavek-kucera: Yes, I'd like to keep it as documentation that those messages existed and optionally that we could reuse them.

SWETAS04 (approved):
Looks good to me .

PR #230, opened 2022-02-22
feat: USING and DROP support

slavek-kucera:
- Implements USING and DROP
- PUSH and POP USING
- Ordinary, labeled, dependent USINGs

- Integration with checker in a separate PR


parser_library/src/context/using.cpp
```diff
@@ -0,0 +1,551 @@
...
+{
+    index_t<instruction_context> ctx_id = add(std::move(eval_ctx), std::move(stack));
+
+    std::vector<index_t<mach_expression>> base;
+    base.reserve(args.size());
+
+    std::transform(args.begin(), args.end(), std::back_inserter(base), [this, ctx_id](auto& a) {
+        return add(std::move(a), ctx_id);
+    });
+    m_usings.emplace_back(current, ctx_id, std::move(base));
```
jirimosinger, line 413: I think `pop_back()` should be used here

parser_library/src/context/using.cpp
```diff
@@ -0,0 +1,551 @@
...
+    dependency_evaluation_context eval_ctx,
+    processing_stack_t stack)
+{
+    index_t<instruction_context> ctx_id = add(std::move(eval_ctx), std::move(stack));
+
+    std::vector<index_t<mach_expression>> base;
+    base.reserve(args.size());
+
+    std::transform(args.begin(), args.end(), std::back_inserter(base), [this, ctx_id](auto& a) {
+        return add(std::move(a), ctx_id);
```
jirimosinger, line 411: I think something like `remove(...)` should be used here

parser_library/src/context/using.cpp
```diff
@@ -0,0 +1,551 @@
...
+
+    auto tmp = get(context_id).context.evaluate(label, owner, offset, long_offset);
+
+    if (tmp.length < 0)
+        return evaluate_result { invalid_register, 1 - tmp.length };
+    else
+        return evaluate_result { tmp.mapping_regs[0], tmp.reg_offset };
+}
+
+template</* std::integral */ typename R, /* std::integral */ typename T>
```
jirimosinger, line 432: Do you want to remove `/* std::integral */`?

parser_library/src/context/using.cpp
```diff
@@ -0,0 +1,551 @@
...
+        clamp<offset_t>(r.min_dist),
+        clamp<offset_t>(ignore_length ? 0 : r.result_entry->length - (offset - r.result_entry->offset)),
+    };
+
+    if (result.reg_offset >= min_disp && result.reg_offset <= max_disp)
+        std::copy(r.result_entry->regs.begin() + r.result_reg, r.result_entry->regs.end(), result.mapping_regs.begin());
+    else if (result.reg_offset < min_disp)
+        result.reg_offset -= min_disp;
+    else /* if (result.reg_offset > max_disp) */
+        result.reg_offset -= max_disp;
```
jirimosinger, line 531: I'm not sure what these two assignments should represent so my question might be irrelevant.

Nevertheless, it seems that when the `result.reg_offset < min_disp` the `min_disp` is subtracted from `result.reg_offset`. Shouldn't it be the other way around? Isn't the end goal to have `result.reg_offset` in the range `<min_disp, max_disp>`?

parser_library/src/context/using.cpp
slavek-kucera: this is actually adding an expression to the vector of expressions, so I think add is appropriate.

parser_library/src/context/using.cpp
slavek-kucera: m_usings vector keeps complete history of all usings/drops and keeps the parent/derived relationship via the m_parent member (here initialized by the current variable).

Eventually, we resolve these relationships step by step - "replaying" the history once everything can be evaluated.

parser_library/src/context/using.cpp
slavek-kucera: I would actually prefer to have it there... but clang-12 is not up to the task yet :(

parser_library/src/context/using.cpp
slavek-kucera: I've been considering adding a union with something like out_of_range_by... because that's what it means when the mapping register set consists of invalid registers.

parser_library/src/expressions/mach_expr_term.h
```diff
@@ -92,7 +100,16 @@ class mach_expr_data_attr final : public mach_expression
 {
     bool do_is_similar(const mach_expression& expr) const override;
 
+    struct private_t
```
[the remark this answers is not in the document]
slavek-kucera: if you can come up with a better way to have a "private" constructor that also works with make_unique I am happy to use it instead of this.

parser_library/src/expressions/mach_operator.cpp
```diff
@@ -85,4 +87,152 @@ size_t mach_expr_unary<par>::hash() const
...
+    auto left_res = left_->evaluate(info);
+    auto right_res = right_->evaluate(info);
+
+    if (!(left_res.value_kind() == context::symbol_value_kind::ABS
+            && right_res.value_kind() == context::symbol_value_kind::ABS)
+        && left_res.value_kind() != context::symbol_value_kind::UNDEF
+        && right_res.value_kind() != context::symbol_value_kind::UNDEF)
+        add_diagnostic(diagnostic_op::error_ME002(get_range()));
+
+    return left_res / right_res;
```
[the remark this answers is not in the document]
slavek-kucera: that should be handled by the operator overload.

parser_library/src/processing/instruction_sets/low_language_processor.cpp
```diff
@@ -51,6 +51,20 @@ context::id_index low_language_processor::find_label_symbol(const rebuilt_statem
...
+    if (const auto& label = stmt.label_ref(); label.type == semantics::label_si_type::ORD)
+    {
+        if (auto [valid, id] =
+                hlasm_ctx.try_get_symbol_name(*std::get<semantics::ord_symbol_string>(label.value).symbol);
+            valid)
+            return id;
+
+        add_diagnostic(diagnostic_op::error_E065(label.field_range));
+    }
+    return nullptr;
```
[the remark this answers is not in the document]
slavek-kucera: nope if the label is missing or is not an ordinary symbol then the using label is simply not there (nullptr).
The diagnostic only indicates that label is present, but does not match the requirements of an ordinary symbol.

parser_library/src/tagged_index.h
```diff
@@ -0,0 +1,49 @@
...
+#define HLASMPLUGIN_PARSERLIBRARY_TAGGED_INDEX_H
+
+#include <cstddef>
+
+namespace hlasm_plugin::parser_library {
+
+template<typename Tag>
+class index_t
+{
+    static constexpr std::size_t invalid_index = (std::size_t)-1;
```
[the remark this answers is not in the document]
slavek-kucera: it is really supposed to be a maximal value of the type - I can replace it with the numeric_limits::max if you prefer.

parser_library/src/expressions/mach_expr_term.cpp
```diff
@@ -85,8 +90,21 @@ context::dependency_collector mach_expr_symbol::get_dependencies(context::depend
 
     if (symbol == nullptr || symbol->kind() == context::symbol_value_kind::UNDEF)
         return value;
+    else if (symbol->kind() == context::symbol_value_kind::ABS && qualifier)
+    {
+        return context::dependency_collector(true);
```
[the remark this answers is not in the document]
slavek-kucera: I think that more extensive overhaul is warranted. so I think I will include it in the next PR...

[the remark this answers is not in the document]
jirimosinger: Ok

parser_library/src/expressions/mach_expr_term.h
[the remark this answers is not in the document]
jirimosinger: Oh, ok. I didn't notice that the constructor is `private`. Unfortunately, I don't have any better/readable alternative

parser_library/src/tagged_index.h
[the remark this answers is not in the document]
jirimosinger: It is just my personal preference to try and assign values of the same datatypes (e.g. non-negative values to unsigned datatypes). Actually, when I'm thinking about it, doesn't Sonar complain about this?

parser_library/src/context/using.cpp
jirimosinger: I like this idea! Maybe it can be a feature for one of the next PRs

jirimosinger (approved):

parser_library/src/tagged_index.h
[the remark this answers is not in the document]
slavek-kucera: I don't see anything in Sonar. And with the explicit cast I would not expect to.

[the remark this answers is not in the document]
jirimosinger: Ok

parser_library/src/context/using.cpp
jirimosinger: Ok

parser_library/src/context/using.cpp
jirimosinger: Ok

parser_library/src/context/using.cpp
jirimosinger: Ok

parser_library/src/expressions/mach_operator.cpp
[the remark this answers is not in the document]
jirimosinger: Ok

Issue #213, opened 2021-12-16
Implement the `&SYSSTMT` system variable


michalbali256:
Implement the `&SYSSTMT` system variable. Specification here:
https://www.ibm.com/docs/en/zos/2.1.0?topic=symbols-sysstmt-system-variable-symbol

PR #234, opened 2022-03-08
feat: Added support for SYSSTMT

jirimosinger:
[Related issue](https://github.com/eclipse/che-che4z-lsp-for-hlasm/issues/213)


parser_library/src/context/hlasm_context.cpp
```diff
@@ -58,6 +58,37 @@ hlasm_context::instruction_storage hlasm_context::init_instruction_map(id_storag
...
+
+class sysstmt_macro_param_data : public macro_param_data_single_dynamic
+{
+    const performance_metrics& metrics;
+    mutable C_t data_ = "";
+
+public:
+    const C_t& get_dynamic_value() const override
+    {
+        size_t sysstmt = metrics.lines + metrics.macro_statements + 1;
```
slavek-kucera, line 79: is this correct even for some of the more obscure cases like macro definition inside a macro call?

parser_library/src/context/hlasm_context.cpp
```diff
@@ -58,6 +58,37 @@ hlasm_context::instruction_storage hlasm_context::init_instruction_map(id_storag
...
+    if (auto value_len = value.size(); value_len < max_number_of_leading_zeros)
+        value.insert(0, max_number_of_leading_zeros - value_len, '0');
+
+    return value;
+}
+
+class sysstmt_macro_param_data : public macro_param_data_single_dynamic
+{
+    const performance_metrics& metrics;
+    mutable C_t data_ = "";
```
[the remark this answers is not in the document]
jirimosinger: I am all for removing the return by reference in favor of return by value.

Nevertheless, this doesn't solve the `mutable` issue as it is needed to overcome having a `const` functions- i.e.:

`virtual C_t get_dynamic_value() const`
and
`C_t get_value() const`

I think I have two options: 
1. I can remove the `const` keyword from all the functions all the way up to the base class. I can also remove `mutable`
2. I can continue to use `mutable` in `get_dynamic_value()`

I am more inclined to go with option `2.` as I feel like the `get_whatever(...)` functions generally shouldn't modify member variables. I also feel like `get_dynamic_value()` is a very special case for which an exception can be made by using `mutable` for a member variable. What do you think?

parser_library/src/context/hlasm_context.cpp
```diff
@@ -58,6 +58,37 @@ hlasm_context::instruction_storage hlasm_context::init_instruction_map(id_storag
...
 namespace {
+template<typename DATA>
+std::string make_string_with_leading_zeros(DATA data, size_t max_number_of_leading_zeros)
+{
+    std::string value = std::to_string(data);
+    if (auto value_len = value.size(); value_len < max_number_of_leading_zeros)
+        value.insert(0, max_number_of_leading_zeros - value_len, '0');
+
+    return value;
+}
```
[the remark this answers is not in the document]
jirimosinger: Agreed

parser_library/src/context/hlasm_context.cpp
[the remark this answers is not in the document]
slavek-kucera: So why not just implement it as
```
C_t get_value() const override
{
   return std::to_string(metrics.lines + metrics.macro_statements + 1);
}
```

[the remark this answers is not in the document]
jirimosinger: Oh, now I understand what you meant by "won't need this mutable storage" 👍

parser_library/src/context/hlasm_context.cpp
jirimosinger: You were right, nested macros have an effect on `SYSSTMT`

parser_library/src/context/hlasm_context.cpp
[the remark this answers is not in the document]
jirimosinger: This PR contains changes that remove the need of having `mutables` and the code does no longer return constant references for values. However, I'm worried a little bit about performance -> I'll run a test after I fix potential issues which might be reported by SonarCloud

parser_library/include/protocol.h
```diff
@@ -361,7 +361,7 @@ struct PARSER_LIBRARY_EXPORT variable
     explicit variable(const debugging::variable& impl);
 
     sequence<char> name;
-    sequence<char> value;
+    std::string value;
```
slavek-kucera, line 364: This should remain non-owning `sequence<char>` "client" view into something that's owned inside the library.

parser_library/src/context/macro_param_data.cpp
```diff
@@ -84,6 +70,21 @@ size_t macro_param_data_composite::size() const { return data_.size(); }
...
+{
+    value_.append("(");
+    for (size_t i = 0; i < data_.size(); ++i)
+    {
+        value_.append(data_[i]->get_value());
+        if (i != data_.size() - 1)
+            value_.append(",");
+    }
+    value_.append(")");
+}
```
slavek-kucera, line 82: Does this work, especially in the light of the new dynamic type?

parser_library/test/debugging/debugger_test.cpp
```diff
@@ -239,13 +240,13 @@ bool check_step(
         if (sc.size() != 3U)
             return false;
         using variables = std::vector<hlasm_plugin::parser_library::variable>;
-        variables globs(d.variables(sc.item(0).variable_reference));
+        variables globs(std::move(d.variables(sc.item(0).variable_reference)));
```
slavek-kucera, line 243: these moves do not make sense to me... `variables(...)` already returns an r-value, doesn't it?

parser_library/src/context/hlasm_context.cpp
jirimosinger: The calculation of SYSSTMT now takes into account also `COPY` statements

parser_library/src/context/macro_param_data.cpp
jirimosinger: It does. Nevertheless, a `debugger` type of test was added to verify this.

parser_library/test/debugging/debugger_test.cpp
jirimosinger: You're right

parser_library/src/context/macro_param_data.cpp
slavek-kucera: But the behavior seems restricted to this one variable...

clients/vscode-hlasmplugin/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 #### Added
 - USING and DROP support
+- SYSSTMT support
+
+#### Fixed
+- Behavior of currently supported subscripted system variables corrected
```
slavek-kucera, line 10: duplicate?

```suggestion
```

slavek-kucera (approved):

PR #235, opened 2022-03-08
refactor: Expressions should not collect diagnostics

slavek-kucera:


parser_library/src/context/ordinary_assembly/dependency_collector.cpp
```diff
@@ -18,42 +18,76 @@
...
 
-dependency_collector& dependency_collector::operator*(const dependency_collector& holder) { return div_mul(holder); }
+dependency_collector& dependency_collector::operator/=(const dependency_collector& holder)
+{
+    if (!merge_undef(holder))
+        div_mul(holder);
 
-dependency_collector& dependency_collector::operator/(const dependency_collector& holder) { return div_mul(holder); }
+    return *this;
+}
```
jirimosinger, line 70: Just making sure - are these two functions supposed to do the same thing? I could imagine that `*=` should behave differently than `/=`

parser_library/src/processing/instruction_sets/asm_processor.cpp
```diff
@@ -130,7 +155,7 @@ void asm_processor::process_EQU(rebuilt_statement stmt)
 
         if (expr_op && !expr_op->has_dependencies(dep_solver))
         {
-            auto length_value = expr_op->expression->resolve(dep_solver);
+            auto length_value = expr_op->expression->evaluate(dep_solver, *this);
```
jirimosinger, line 158: It feels a little bit weird to provide `*this` for the `diagnostic_op_consumer...` in call to `evaluate(...)`.

Is it possible/meaningful to not have this inheritance relationship on `diagnosable_ctx` and have it in a form of a member variable instead?

jirimosinger (approved):

parser_library/src/processing/instruction_sets/asm_processor.cpp
slavek-kucera: eventually yes, but the PR seems quite big already, so I thought leaving it for another PR in the future... at least there is only handful of processors unlike thousands of expressions.

Issue #236, opened 2022-03-14
AINSERT with apostrophes


jirimosinger:
**Describe the bug**
An error occurs when using `AINSERT` in combination with e.g. `''&SYSSTMT'''`

**To Reproduce**
```
    MACRO
    MAC
    AINSERT '&&B    SETC ''&SYSSTMT''',BACK
    MEND
    
    MAC
    
    END
```

**Expected behavior**
`&B` is assigned value `00000007` without errors

**Screenshots**
![image](https://user-images.githubusercontent.com/99467904/158218057-a3d2790a-6c57-4394-ab11-bff11e3831c9.png)

**Platform**
 - Editor: VS Code 1.65.2
 - OS: Windows
 - Plugin version: 1.0.0

PR #243, opened 2022-03-17
fix: AINSERT immediate variable evaluation

jirimosinger:
- AINSERT specific tests were moved to a separate test file
- See [related issue](https://github.com/eclipse/che-che4z-lsp-for-hlasm/issues/236)


parser_library/src/parsing/grammar/model_operand_rules.g4
```diff
@@ -57,14 +57,21 @@ model_string returns [std::string value]
 		collector.add_hl_symbol(token_info(provider.get_range($ap1,$ap2),hl_scopes::string));
 	};
 
+before_var_sym_model_string returns [std::string value]
+	: ap1=APOSTROPHE model_string_ch_c ap2=(APOSTROPHE|ATTR)
+	{
+	    $value.append(std::move($model_string_ch_c.value));
```
slavek-kucera, line 63: does this perform the `if ($l_sp_ch.value == "&&")` replacement? Or is it not needed?

jirimosinger: No, during the first passthrough it doesn't do the `&&` replacement. Actually, one could see the replacement to be part of the issue.

There are two passthroughs that happen when strings like `AINSERT '&&B    SETC ''&SYSSTMT''',BACK` are processed.

Originally, during the first passthrough `ANTLR` would return the following result `AINSERT '&B    SETC ''00000009''',BACK`. Unfortunately, during the second pass the string would go through the same branch as before which ultimately resulted in error.

The idea behind this modification is that the processing during the second pass should behave similarly to the processing of a simpler string like `AINSERT '&&B    SETC ''00000009''',BACK` which requires only 1 passthrough. That means that only the `&SYSSTMT` part of the original string is modified and the rest stays the same until it go to `ANTLR` for the second time.

slavek-kucera: 👍 
I suggest adding additional tests with multiple substitutions and `''` sequences...

PR #255, opened 2022-03-30
feat: DB2 preprocessor VERSION option

slavek-kucera:


parser_library/src/processing/preprocessors/db2_preprocessor.cpp
```diff
@@ -70,8 +72,42 @@ class db2_preprocessor : public preprocessor
...
+        if (m_version.size() <= version_chunk)
+        {
+            m_buffer.append("SQLVERSP DC    CL4'VER.' VERSION-ID PREFIX\n");
+            m_buffer.append("SQLVERD1 DC    CL64'").append(m_version).append("'        VERSION-ID\n");
+        }
+        else
+        {
+            m_buffer.append("SQLVERS  DS    CL68      VERSION-ID\n"
+                            "         ORG   SQLVERS+0\n"
+                            "SQLVERSP DC    CL4'VER.' VERS-ID PREFIX\n");
```
jirimosinger, line 89: I just want to check that it is OK to use `VERS-ID` instead of `VERSION-ID` which is used in all the lines above.

And basically the same question applies to line 99 where `VERS-ID` is used as well

jirimosinger: Cool 👍

jirimosinger (approved):

jirimosinger (approved):

PR #270, opened 2022-06-10
Update README.md

zeibura:
We need to add a new section to the readme for Zowe v2 compliance recommending the use of Zowe Explorer with this extension. I can only think of one benefit, which is opening files from the mainframe, if there are any others let me know.

Signed-off-by: Zeibura Kathau <zeibura.kathau@broadcom.com>

### DIFF
--- a/clients/vscode-hlasmplugin/README.md
+++ b/clients/vscode-hlasmplugin/README.md
@@ -21,6 +21,12 @@ HLASM Language Support is also part of [Code4z](https://marketplace.visualstudio
 
 HLASM Language Support is supported on Visual Studio Code and Github Codespaces.
 
+## Integration with Zowe Explorer
+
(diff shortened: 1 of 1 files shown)


clients/vscode-hlasmplugin/README.md
```diff
@@ -21,6 +21,12 @@ HLASM Language Support is also part of [Code4z](https://marketplace.visualstudio
 
 HLASM Language Support is supported on Visual Studio Code and Github Codespaces.
 
+## Integration with Zowe Explorer
+
+We recommend that you use HLASM Language Support with Zowe Explorer. Zowe Explorer enables you to open your High Level Assembler programs stored on mainframe data sets directly in VS Code.
```
slavek-kucera, line 26: Shouldn't we first verify that this is actually working reasonably well?

zimlu02: Screenshot 2022-06-10 at 16 39 06

zimlu02: looks good to me - what do you want to test?

